Issue #10, opened 2018-07-20
Design, implement, test extract config python module and ExtractConfig class


znatty22:
Implement the ExtractConfig class which will load the extract config Python module, validate it, and deserialize it into the ExtractConfig class.

Write unit tests to ensure validation and deserialization work.

PR #56, opened 2018-09-18
✨ Extract config design + basic operations

fiendish:
Should resolve https://github.com/kids-first/kf-lib-data-ingest/issues/10


src/etl/extract/operations.py
```diff
@@ -0,0 +1,178 @@
...
+    dataframe as input and returns a new dataframe.
+
+    :param m: A function to apply to a DataFrame
+    :returns: A function that applies the specified m operation
+    """
+    assert_safe_type(m, function)
+
+    def df_map_func(df):
+        new_df = df.copy()
+        return m(new_df)
```
znatty22: We should probably do an assert safe type on the output of m. It should be a dataframe

fiendish: you did it! 🙆‍♂️

PR #73, opened 2018-10-26
✨ Extract pipeline

fiendish:
branched from #56. please review that first and then I'll rebase.


znatty22:
We should add more tests for the extract stage, but we can do that in another PR. Let's get in what we have right now

PR #77, opened 2018-11-02
 ✅ more tests for extract output

fiendish:
also adds the keep_map shortcut operation

PR #79, opened 2018-11-09
✨ serialize/deserialize extract stage output

fiendish:
branched from extract-testing (PR #77). approve that first, please. will rebase when that merges.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,6 @@ venv.bak/
 
 # Visual studio
 .vscode/
+
+# stage serialization output
+*_cache.txt
--- a/src/common/misc.py
+++ b/src/common/misc.py
@@ -55,7 +55,7 @@ def intsafe_str(val):
     Returns:
         string: Representation of `val`
     """
-    if isinstance(val, float) and numpy.isnan(val):
+    if val is None or (isinstance(val, float) and numpy.isnan(val)):
         return None
     val = str(val).strip()
     # I hate PEP 515.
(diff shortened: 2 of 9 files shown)


fiendish:
@znatty22 
> In #68 I was thinking that we shouldn't force a stage to write all of its output to a single file. Instead we allow the stage's to write_output to write output however it would like into an output dir. Then we leave it up to the stage's read_input to be responsible for understanding the output file(s) to read in as input.

In that case we should collapse write/serialize read/deserialize into 2 functions instead of 4.

> In this PR I see that the extract stage now needs to know the study_config_dir (aka dir of dataset ingest config). I think we can avoid this by allowing ingest_pipeline to construct the default output dir for a stage based on the dataset ingest config dir

This sounds reasonable. 

Can you detach stage serialization from #68 and merge it into this branch?

znatty22:
@fiendish 
> In that case we should collapse write/serialize read/deserialize into 2 functions instead of 4.
Yep that's what I've done in #68

>Can you detach stage serialization from #68 and put it into this branch so we can work on it here instead of a multi-feature monolith?

Hmm it would probably be better to branch off master, make a simple PR for the refactor and then merge that in. Then you should be able to just rebase on master since it looks like this branch is already up to date w master

fiendish:
@znatty22 
> In this PR I see that the extract stage now needs to know the study_config_dir (aka dir of dataset ingest config). I think we can avoid this by allowing ingest_pipeline to construct the default output dir for a stage based on the dataset ingest config dir

What's the difference? In both cases you need to pass a directory to the stage, otherwise it won't know where to write to.

fiendish:
@znatty22 I think 5ec662a is easier

PR #84, opened 2018-11-16
🐛 use auth profiles for s3 requests

fiendish:
Use of the saml profile was being overridden by None. I guess the s3 file retriever test doesn't mock requiring authentication. Anyway, this should fix it. Boyd test files loaded from s3 verify this.

`kidsfirst ingest tests/data/test_boyd_study/` successfully extracts now using the remote files

### DIFF
--- a/src/common/stage.py
+++ b/src/common/stage.py
@@ -7,8 +7,8 @@ from functools import wraps
 
 class IngestStage(ABC):
 
-    def __init__(self, study_config_dir=None):
-        self.study_config_dir = study_config_dir
+    def __init__(self, stage_cache_dir=None):
+        self.stage_cache_dir = stage_cache_dir
         self.logger = logging.getLogger(type(self).__name__)
 
     @abstractmethod
@@ -22,7 +22,7 @@ class IngestStage(ABC):
         pass
 
     @abstractmethod
-    def _serialize_output(self, output):
+    def _write_output(self, output):
         """
         An ingest stage is responsible for serializing the data that is
         produced at the end of stage run.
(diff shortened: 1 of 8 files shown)


src/common/file_retriever.py
```diff
@@ -35,7 +37,7 @@ def split_protocol(url):
         return None, None
 
 
-def _s3_save(protocol, source_loc, dest_obj, auth="saml"):
+def _s3_save(protocol, source_loc, dest_obj, auth, logger):
```
znatty22: Couldn't auth still be a kwarg? It looks like you expect it to be None in some cases anyway?

src/common/file_retriever.py
```diff
@@ -45,17 +47,31 @@ def _s3_save(protocol, source_loc, dest_obj, auth="saml"):
...
+            s3.Object(bucket, key).download_fileobj(dest_obj)
+            return
+        except botocore.exceptions.NoCredentialsError:
+            pass
+
+    raise botocore.exceptions.NoCredentialsError()  # never got the file
 
-def _web_save(protocol, source_loc, dest_obj, auth=None):
+
+def _web_save(protocol, source_loc, dest_obj, auth, logger):
```
znatty22: Since requests.get takes auth as a kwarg, couldn't it be a kwarg in _web_save too?

src/common/file_retriever.py
fiendish: You can always pass it as a kwarg anyway even if it isn't defined as one. Anyway the internal calls are all underscores, and the get call has it as a kwarg. I don't know. Yes? 😆 But for what?

znatty22: Well I guess if these individual save methods are never going to be used independently of the file retriever then it doesn't matter. If they are, then auth should be optional. I'm ok with it the way it is

PR #86, opened 2018-11-19
 🐛 continue past failing s3 profiles

### LINKED ISSUE #85: file retriever s3 save fails on first invalid credentials
_s3_save will cycle through the available profiles in a ~/.aws/credentials file,  and try each one until it finds one that works or raise an exception if it can't find one. 

If you have multiple profiles in your credentials file, _s3_save may potentially try one that fails authentication. This raises a `botocore.exceptions.ClientError`. Should probably change the method to catch both `botocore.exceptions.NoCredentialsError` and `botocore.exceptions.ClientError`.

Unfortunately the `botocore.exceptions.ClientError` isn't very descriptive. It just has a 400 status code with a Bad Request message. It would have been nicer to catch `botocore.exceptions.ClientError` and if we know its an auth failure, then we pass the exception so _s3_save may keep trying credentials.

fiendish:
should fix #85

### DIFF
--- a/src/common/file_retriever.py
+++ b/src/common/file_retriever.py
@@ -66,7 +66,12 @@ def _s3_save(protocol, source_loc, dest_obj, auth=None, logger=None):
             s3 = boto3.session.Session(profile_name=profile).resource("s3")
             s3.Object(bucket, key).download_fileobj(dest_obj)
             return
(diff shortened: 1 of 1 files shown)


fiendish:
I think https://github.com/boto/botocore/issues/1400 is related to our ability to handle this. Let's keep an eye on it.

PR #88, opened 2018-11-20
📦 Fix package installation

### LINKED ISSUE #87: Modify installation of kf-lib-data-ingest so that it has better namespace
**Problem**
Currently when you install the kf-lib-data-ingest lib, the src/ is add to sys path and everything under that directory is made available for import. If you want to use this as an actual library, this is problematic since the modules in the src/ dir are named very generically. 

For example, I want to use the kf-lib-data-ingest etl.extract package in my OMOP ETL code
```
.
├── code
│   ├── extract.py
│   ├── load.py
│   ├── etl.py
│   └── transform.py
```
```python
# My extract.py file
from etl.extract.extract import ExtractStage
def run(output_dir, extract_configs):
    es = ExtractStage(output_dir, extract_configs)
    df_out = es.run()
    es._write_output(df_out)

    return df_out
```
But I get namespace collisions with my etl.py module

**Solution**
Change setup.py to find packages in the top level repo dir rather than the src dir. Imports will then become: 

```python
from kf_lib_data_ingest.etl.extract import ExtractStage
```

znatty22:
We want to be able to use this as a library. Change the package installation so that all modules are under a root package - kf_lib_data_ingest.

Fixes #87


znatty22:
@fiendish Sorry its probably autopep8 doing its formatting, i'll fix

fiendish:
The unindents are a bit of an odd situation. We could also assign those path joins to temporary variables to make the table structure less funky.

tests/test_extract_stage.py
```diff
@@ -40,7 +39,7 @@ def test_extracts():
         for config in extract_configs:
             extracted = df_out[config][1]
             expected = es._source_file_to_df(
-                "file://" + study_configs[config]
+                "file://"+study_configs[config]
```
fiendish: ok but un-undo this one 😄

fiendish:
🚚 🚚 🚚 🚚 🚚 🚚 🚚 🚚 🚚 🚚

PR #93, opened 2018-12-18
✨ support splitting into multiple rows

### LINKED ISSUE #53: Operation: split rows into multiple rows
Would be nice to have a helper function to split rows into multiple rows based on comma separated values occurring in a column of choice. 

For example, Onel 2015 has a lot of data like this: 

| tumor_location_uberon | foo |
| ----------------------- | ----|
| UBERON:0002168, UBERON:0002167, UBERON:0003728 | A

I want a function/operation that does:
```python
def split_rows(df, col, delimiter):
    # split rows with delimited string into individual rows
    return new_df

new_df = split_rows(df, 'tumor_location_uberon', ',')
```
and results in:

| tumor_location_uberon | foo |
| ----------------------- | ---- |
| UBERON:0002168 | A
| UBERON:0002167 | A
| UBERON:0003728 | A

fiendish:
closes #53


kf_lib_data_ingest/etl/extract/extract.py
```diff
@@ -7,17 +8,14 @@
...
-    split_protocol
-)
+from kf_lib_data_ingest.common.file_retriever import (PROTOCOL_SEP,
+                                                      FileRetriever,
+                                                      split_protocol)
 from kf_lib_data_ingest.common.misc import intsafe_str
 from kf_lib_data_ingest.common.stage import IngestStage
 from kf_lib_data_ingest.common.type_safety import function
-from kf_lib_data_ingest.etl.configuration.base_config import (
+from kf_lib_data_ingest.etl.configuration.base_config import \
```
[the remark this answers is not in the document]
fiendish: nothing's wrong with the other way. this is just what my auto-import-formatter does

fiendish:
> I don't think we should blindly split rows based on columns that lists. I think the user should specify which columns they want to split by, and what delimiters to use.

Can you think of a reason why you would otherwise want to return a list of values for a single cell? I can't.

> I think the user should specify which columns they want to split by, and what delimiters to use.

Delimiters can be arbitrarily complex and arbitrarily different from cell to cell, so I don't think that would work as well.

kf_lib_data_ingest/etl/extract/extract.py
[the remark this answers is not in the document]
fiendish: I can customize it to use any of https://github.com/timothycrosley/isort#multi-line-output-modes

My preference is for 5

fiendish:
> a method in kf_lib_data_ingest.common.misc... This way its standalone and can be used outside of extract.py too.

We can do that too by moving split_row_lists into common and wrapping it as appropriate.

kf_lib_data_ingest/common/pandas_utils.py
```diff
@@ -6,9 +6,81 @@
 import numpy
 import pandas
 
+from kf_lib_data_ingest.common.misc import multisplit
 from kf_lib_data_ingest.common.type_safety import assert_safe_type
 
 
+class Split:
```
znatty22: jw why you didn't do:

```python
class Split:
    def __init__(self, things):
        assert_safe_type(things, list)
        self.things = things

s = Split() 
s.things
```

kf_lib_data_ingest/etl/extract/extract.py
```diff
@@ -172,11 +169,21 @@ def _run(self):
...
             # extraction
             df_out = self._chain_operations(
                 df_in, extract_config.operations
             )
 
+            # split value lists into separate rows
+            df_out = split_df_rows_on_splits(
+                                                df_out.reset_index()
+                                            ).set_index('index')
+            del df_out.index.name
```
znatty22: Why is this needed?

kf_lib_data_ingest/common/pandas_utils.py
fiendish: I originally had a different vision for it and this is just leftover. No reason now to not look at things directly.

kf_lib_data_ingest/etl/extract/extract.py
fiendish: Having a named index just offends me. No other reason.

kf_lib_data_ingest/common/pandas_utils.py
fiendish: 👍

PR #57, opened 2018-09-27
✨ Standard model + transformation

znatty22:
Addresses #25

PR #98, opened 2019-01-16
♻️ Minor refactor transform, load

znatty22:
Part 1 of #95 

Previously the load stage was invoking a part of the transformation (from concept graph to target entities). We want to move this responsibility into the transformation stage and ensure that the load stage only deals with transmitting the target entities coming out of transform stage into the target service. 

This will allow us to interchange the transformation approach at runtime (auto transform via graph or guided transform via user supplied transform function).

### DIFF
--- a/tests/test_standard_model.py
+++ b/tests/test_standard_model.py
@@ -123,103 +123,103 @@ def random_model(random_data_df_dict):
     return model
 
 
-# def test_populate_model(data_df_dict):
-#     """
-#     Test populating the concept graph in the standard model
-#     """
-#     model = StandardModel()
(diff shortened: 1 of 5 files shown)


tests/test_standard_model.py
```diff
@@ -123,103 +123,103 @@ def random_model(random_data_df_dict):
     return model
 
 
-# def test_populate_model(data_df_dict):
```
znatty22: Can't believe I didn't notice this in #57, but I commented some tests out while debugging other tests and forgot to come back and uncomment 😬

fiendish:
part 1 of 95! 😮 94 more parts to go!

PR #100, opened 2019-01-17
♻️ Refactor transform stage to use Auto/Guided transformers

znatty22:
In preparation for implementing the guided transformer, we need to: 

- [x] Define two new classes to encapsulate the two transformation approaches: auto transformation (via graph) and guided transformation (via user supplied transform function).  
- [x] Refactor the transform stage so that upon init, it can accept a CLI parameter that specifies which transformer to use and then invoke the desired transformer
- [x] Move the unique key column insertions, previously happening in the standard model's transform method, into the top level transform stage.


kf_lib_data_ingest/etl/transform/standard_model/model.py
```diff
@@ -3,13 +3,6 @@
 import pandas as pd
 
 from kf_lib_data_ingest.etl.transform.standard_model.graph import ConceptGraph
-from kf_lib_data_ingest.etl.transform.standard_model.concept_schema import (
```
znatty22: Unique key column generation was previously being done in the standard model, but we will want this regardless of which transformation approach is used.  

** Quick refresher, unique key generation is the process of explicitly specifying the column that should be used to uniquely identify a concept. For most concepts, its just the ID (CONCEPT.PARTICIPANT.ID) but for some, like diagnosis, it would be a combination of other columns.

tests/conftest.py
```diff
@@ -60,3 +64,8 @@ def ingest_pipeline():
 @pytest.fixture(scope='function')
 def target_api_config():
     return TargetAPIConfig(KIDS_FIRST_CONFIG)
+
+
+@pytest.fixture(scope='function')
```
znatty22: Make a reusable fixture for transform stage because now we need to use it across multiple tests

tests/test_standard_model.py
```diff
@@ -371,107 +375,3 @@ def test_gml_import_export(tmpdir_factory, model):
         assert node
         assert (ConceptNode.to_dict(node['object']) ==
                 ConceptNode.to_dict(orig_node['object']))
-
```
znatty22: These tests having to do with unique key stuff have been moved into `test_transform_stage.py` since standard model no longer does unique key gen.

tests/test_transform_stage.py
```diff
@@ -1,25 +1,134 @@
...
-        stage.run({i: 'foo' for i in range(5)})
+        transform_stage.run({i: 'foo' for i in range(5)})
 
     # Bad values
     with pytest.raises(InvalidIngestStageParameters):
-        stage.run({'foor': ('bar', None) for i in range(5)})
+        transform_stage.run({'foor': ('bar', None) for i in range(5)})
+
+
+def test_unique_keys(transform_stage):
```
znatty22: Unique key gen tests have been moved from `test_standard_model.py` to here

kf_lib_data_ingest/etl/transform/auto.py
```diff
@@ -0,0 +1,58 @@
...
+    assert_safe_type,
+    assert_all_safe_type
+)
+from kf_lib_data_ingest.etl.transform.standard_model.model import StandardModel
+from kf_lib_data_ingest.etl.configuration.target_api_config import (
+    TargetAPIConfig
+)
+
+
+class AutoTransformer(IngestStage):
```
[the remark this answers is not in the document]
znatty22: > Alternatively, these classes don't need to exist. Just replace TransformStage's main function with one loaded from the config file. Inputs and outputs should be identical between transform functions.

Can't do that. GuidedTransform is going to do more than applying the user's transform funct to merge tables together. It will also include converting the merged tables into target entity dicts (CONCEPT.SPECIMEN.ANALYTE -> biospecimen.analyte_type, etc).  

> This is either defining a pipeline stage or not.
True, these no longer need to be ingest stages. They could just be submodules. Hmm I think I like the submodule approach better than the inheritance approach. I think `ingest_pipeline.py` should be kind of be "dumb" and focus on a few simple things: know the order in which to run stages, and know which params get passed on to which stages, and executing a stage via its run method. The stages themselves should take in their input params and know how to interpret them to do things like load submodules.

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -92,16 +93,33 @@ def _run(self, target_api_config_path, use_async=False,
...
                                 self.data_ingest_config.extract_config_paths)
-        self.stage_dict['t'] = (TransformStage, target_api_config_path)
 
+        # Transform stage
+        transform_funct_path = self.data_ingest_config.transform_function_path
+
+        if transform_funct_path:
+            # User only supplied filename, assume file is in ingest_config_dir
+            dirpath, filename = os.path.split(transform_funct_path)
+            if not dirpath:
```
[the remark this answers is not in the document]
znatty22: ok 👍

kf_lib_data_ingest/etl/transform/guided.py
```diff
@@ -0,0 +1,57 @@
...
+        pass  # TODO
+
+    def _validate_run_parameters(self, data_dict):
+        """
+        Implemented by TransformStage
+        """
+        pass
+
+    def _run(self, data_dict):
+        """
```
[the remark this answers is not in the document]
znatty22: Will be implemented in #101. Just needs to return something to complete the pipeline for now

kf_lib_data_ingest/etl/transform/auto.py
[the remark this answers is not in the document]
fiendish: > Can't do that. GuidedTransform is going to do more than apply the user's transform funct to merge tables together. It will also include converting the merged tables into target entity dicts (CONCEPT.SPECIMEN.ANALYTE -> biospecimen.analyte_type, etc).

So what? That's all part of TransformStage._run.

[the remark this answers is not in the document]
znatty22: Let's discuss IRL lol

kf_lib_data_ingest/etl/transform/auto.py
```diff
@@ -0,0 +1,35 @@
...
+source data tables.
+"""
+import logging
+
+from kf_lib_data_ingest.etl.transform.standard_model.model import StandardModel
+
+
+class AutoTransformer():
+    def __init__(self, target_api_config):
+        self.logger = logging.getLogger(type(self).__name__)
```
fiendish: Not for this PR, but it looks like we have a couple of these invocations in the code. Maybe we should make a default logger mixin?

kf_lib_data_ingest/etl/transform/guided.py
```diff
@@ -0,0 +1,31 @@
...
+should be merged in order to yield a single table per target service entity.
+"""
+import logging
+
+from kf_lib_data_ingest.etl.configuration.transform_module import (
+    TransformModule
+)
+
+
+class GuidedTransformer():
```
fiendish: I still don't understand why GuidedTransformer and AutoTransformer are separate classes instead of one class that optionally uses an external module to override some or all of its run method during init.

PR #107, opened 2019-02-04
✨ preserve row index across extract

fiendish:
put the row index that a value came from into the extract output

### DIFF
--- a/kf_lib_data_ingest/common/datafile_readers.py
+++ b/kf_lib_data_ingest/common/datafile_readers.py
@@ -30,4 +30,4 @@ def read_excel_file(filepath_or_buffer, **kwargs):
             wb = xlrd.open_workbook(
                 filename=filepath_or_buffer, logfile=devnull
             )
-    return pandas.read_excel(wb, engine='xlrd', **kwargs)
+    return pandas.read_excel(wb, engine='xlrd', dtype=object, **kwargs)
--- a/kf_lib_data_ingest/common/datafile_readers.py
+++ b/kf_lib_data_ingest/common/datafile_readers.py
@@ -30,4 +30,6 @@ def read_excel_file(filepath_or_buffer, **kwargs):
             wb = xlrd.open_workbook(
                 filename=filepath_or_buffer, logfile=devnull
             )
-    return pandas.read_excel(wb, engine='xlrd', dtype=object, **kwargs)
+    kwargs['engine'] = 'xlrd'
+    kwargs['dtype'] = object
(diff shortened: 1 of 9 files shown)


kf_lib_data_ingest/etl/extract/extract.py
```diff
@@ -270,5 +292,13 @@ def _chain_operations(self, df_in, operations):
                 df_out[col_name] = pandas.Series(
                     list(col_series) * round(length_multiplier)
                 )
-
+                col_index = list(col_series.index) * round(length_multiplier)
```
znatty22: In the example, commented above the for loop, can you add in the index col in the before and after? This doesn't seem like it will work but I'm not sure what the end result is supposed to look like in terms of the index.

fiendish: Indexes don't actually apply here except as a consistency check. This scenario only applies to melt-like operations where the number of rows in a given column will be an exact multilpe of the number of rows in the shortest column. For this example the indices are all 0.

kf_lib_data_ingest/common/datafile_readers.py
```diff
@@ -30,4 +30,4 @@ def read_excel_file(filepath_or_buffer, **kwargs):
             wb = xlrd.open_workbook(
                 filename=filepath_or_buffer, logfile=devnull
             )
-    return pandas.read_excel(wb, engine='xlrd', **kwargs)
+    return pandas.read_excel(wb, engine='xlrd', dtype=object, **kwargs)
```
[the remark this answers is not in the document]
fiendish: will do

kf_lib_data_ingest/etl/extract/extract.py
fiendish: Alternate example:

```Python
        # Given...
        #
        #  index   A   B   C
        #  0       1   3   5
        #  1       2   4   6
        #
        # ... (with B and C each being melted) will temporarily extract to ...
        #
        #  A_index A   DESCRIPTION_index   DESCRIPTION   VALUE_index   VALUE
        #  0       1   0                   B             0             3
        #  1       2   1                   B             1             4
        #              0                   C             0             5
        #              1                   C             1             6
        #
        # ... and will then fill out to ...
        #
        #  A_index A   DESCRIPTION_index   DESCRIPTION   VALUE_index   VALUE
        #  0       1   0                   B             0             3
        #  1       2   1                   B             1             4
        #  0       1   0                   C             0             5
        #  1       2   1                   C             1             6
```

PR #110, opened 2019-02-12
💄 Update docs to use Kids First Style

znatty22:
- Use the standard Kids First style sheet instead of default sphinx readthedocs style
- Fix obsolete source code paths

Looks like this:
image


znatty22:
@dankolbman We may want to consider putting the style sheet along with other common Kids First assets into a central place, perhaps a git repo, idk? We're using it in two places now and I can see us using it on other repos with sphinx docs

dankolbman:
Yeah, I'm not sure how to organize it best. We also use it in airflow docs now, too.
We could put the stylesheet up as a static file and pull it into the page, but we'd have to require all docs sites to link it in their templates.

fiendish:
All files on github are available in static format via raw.githubusercontent.com

dankolbman:
@fiendish yeah that's true. Maybe we should just link in the developer handbook style sheet on all other docs sites?

fiendish:
github also supports dependent sub-repositories

znatty22:
@dankolbman @fiendish There doesn't seem to be a simple, one-liner approach to link in an external style sheet. Since `conf.py` is a python module we could probably write something in there to download and overwrite the current `style.css` to the latest one from somewhere. 

But, maybe there is a better approach. We could probably have a separate repo (i.e. sphinx_kidsfirst_theme) for a Sphinx docs Kids First theme. Then instead of having `sphinx_rtd_theme` in `requirements.txt` we have our sphinx_kidsfirst_theme in there.  Doesn't look too difficult according to: http://www.sphinx-doc.org/en/stable/theming.html.

I'll try it out real quick to see if its really as straight forward as it seems. If so then I can update the requirements in this PR. If not, then I'll merge this PR in as-is and create an issue to figure this out in the near future.

znatty22:
TODO: Use KF sphinx theme

znatty22:
It shouldn't. That last commit fixes it!

PR #119, opened 2019-02-14
👷 Fix CI Scripts

znatty22:
- For all CI scripts, always abort on non-zero exit code from any of the things running in the script
- Upgrade pip and install egg info before installing reqs
    - See https://github.com/pypa/pip/issues/4537 for more details
- Fix recurring S3 related error in file retriever test

thanks @fiendish !


fiendish:
Ok, I fixed circle. Good luck, though. Now you can't merge until the tests pass!

PR #128, opened 2019-02-18
📝 add row-splitting to the tutorial

fiendish:


docs/source/tutorial/extract.rst
```diff
@@ -345,6 +350,86 @@ The resulting intermediate output will look like:
...
+    )
+
+This says take "Use the ``specimens`` column as input, split any ``,`` or ``;``
+delimited values apart into their own entries, and then output the result to a
+standard concept column for biospecimen ID."
+
+.. note::
+
+    We use a special ``Split()`` object for lists of values that we want to
+    split apart. That lets us preserve the ability to return a list of values
```
znatty22:
```suggestion
    split apart into multiple rows. We could have split the string into a list and then used the existence of a list as the signal to split a row into multiple rows. However, we wanted to leave the option for a cell value to contain a list. This is why we use the Split object, which explicitly tells the extract stage to split a list of values into multiple rows.
```

docs/source/tutorial/extract.rst
```diff
@@ -345,6 +350,86 @@ The resulting intermediate output will look like:
...
+    pertaining to a single entity without splitting them apart if we need to.
+
+.. note::
```
znatty22: Not sure if this note is necessary

docs/source/tutorial/extract.rst
```diff
@@ -345,6 +350,86 @@ The resulting intermediate output will look like:
...
+
+.. code-block:: python
+
+    value_map(
+        in_col="specimens",
+        m=lambda x: Split(re.split('[,;]', x)),
+        out_col=CONCEPT.BIOSPECIMEN.ID
+    )
+
+This says take "Use the ``specimens`` column as input, split any ``,`` or ``;``
```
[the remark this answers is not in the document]
fiendish: 👌

docs/source/tutorial/extract.rst
fiendish: 〰️

docs/source/tutorial/extract.rst
fiendish: 👍

PR #138, opened 2019-02-20
🐛 Fix overwrite log issues

### LINKED ISSUE #137: Fix overwrite_log issues
1. Overwrite the ingest log by default

    The default is `overwrite_log=False` which causes an `ingest-<timestamp>.log` to be written every 
    time I do `kidsfirst ingest <study_dir>`. The default should be `overwrite_log=True` so that 
    `ingest.log` is generated once and updated on subsequent runs

2. `overwrite_log` has wrong type in dataset_ingest_config.yaml schema

    Currently the type is set to string, it should be boolean

znatty22:
Fixes #137


tests/test_logging.py
```diff
@@ -31,16 +31,10 @@ def test_defaults(ingest_pipeline):
...
             DEFAULT_LOG_OVERWRITE_OPT)
 
-    # Test that ingest.log was not created but ingest-{datetime}.log
-    # was created
+    # Test that ingest.log was created
     ingest_pipeline.run(target_api_config_path)
     log_filepath = os.path.join(default_log_dir, DEFAULT_LOG_FILENAME)
-    assert os.path.isfile(log_filepath) is False
-
-    # Test that new logs are created on each run
```
fiendish: Where did this test go?

fiendish: oh nvm .lol

PR #143, opened 2019-02-22
 🐛 normalize dtypes in extract output to string

fiendish:


fiendish:
During accounting I noticed that I was seeing both (e.g.) 1 and '1' as values in the outputs from extract.

kf_lib_data_ingest/etl/extract/extract.py
```diff
@@ -184,6 +184,10 @@ def _run(self):
                                             ).set_index('index')
             del df_out.index.name
 
+            # standardize on string values
```
znatty22: So this is converting all non-nulls to strings? What's the reason for doing this?

This might affecting loading later on. I think the validation layer (before the json gets deserialized) in the dataservice expects certain types for various columns

fiendish: > So this is converting all non-nulls to strings?

Yes. 

> What's the reason for doing this?

Because I believe that it's semantically the correct thing to do, since cells don't ever contain quotation marks. You would never choose to differentiate between PID="1" and PID=1.

> This might affecting loading later on.

I don't think it will, but if it does then that should be considered a dataservice bug or possibly a loader bug. We have more important needs than the dataservice deciding to be inflexible for no reason.

znatty22: Ok 👍, then the loader will need to know what the dataservice needs which it can do by looking at the schemas from `https://kf-api-dataservice.kidsfirstdrc.org/swagger`. 

Just an FYI - this matters when we are processing nulls at the end of transform or beginning of load (wherever we decide). Null numeric types in a DataFrame are NaNs, null string and boolean types are None. At the end of transform I need to convert any null strings to `Not Reported` and any null numeric or boolean types to None. That's why I need to know the types of each field. 

Example transform output, before null processing happens:
```
{
            "id":"5136-SB-1112",
            "links":{
                "participant_id":"10000-1",
                "sequencing_center_id":null
            },
            "properties":{
                "age_at_event_days":null,
                "analyte_type":null,
                "composition":NaN,
                "concentration_mg_per_ml":null,
                "external_aliquot_id":"5136-SB-1112",
                "external_sample_id":"5136-SB-1112",
                "ncit_id_anatomical_site":null,
                "ncit_id_tissue_type":null,
                "shipment_date":null,
                "shipment_origin":null,
                "source_text_anatomical_site":NaN,
                "source_text_tissue_type":"Normal",
                "source_text_tumor_descriptor":null,
                "spatial_descriptor":null,
                "visible":null,
                "volume_ml":null
            }
        }
```
Example transform output, after null processing happens:
```
        {
            "id":"5136-SB-1112",
            "links":{
                "participant_id":"10000-1",
                "sequencing_center_id":None
            },
            "properties":{
                "age_at_event_days": None,
                "analyte_type": "Not Reported",
                "composition": None,
                "concentration_mg_per_ml": None,
                "external_aliquot_id":"5136-SB-1112",
                "external_sample_id":"5136-SB-1112",
                "ncit_id_anatomical_site":"Not Reported",
                "ncit_id_tissue_type":"Not Reported",
                "shipment_date": None,
                "shipment_origin":"Not Reported",
                "source_text_anatomical_site": "Not Reported",
                "source_text_tissue_type": "Normal",
                "source_text_tumor_descriptor": "Not Reported",
                "spatial_descriptor": "Not Reported",
                "visible": None,
                "volume_ml": None
            }
        }
```

fiendish: > Null numeric types in a DataFrame are NaNs

That doesn't sound quite right. NaN is specifically a float type, not all numerics. Mixing None with ints requires either conversion to floats, which converts None to NaN, or mixed object dtype which leaves the Nones alone. Because pandas has an int dtype but nan is a float, numbers get treated differently when loading from a file depending on whether there are missing values in the column or not, which introduces inconsistent behavior. That's why all of the data loading by extract is done with dtype=object specifically so that values that sort of maybe look like numbers don't sometimes get automatically coerced into different types and sometimes not.

znatty22: Ok that makes sense. Either way, transform will need to know the types of columns when converting null values to the appropriate target value.

PR #145, opened 2019-02-25
♻️ Natasha PR Review Accounting

znatty22:
Few suggestions:
1.  Break up the `_post_extract` method into modular components (tests) that can be loaded on a per study basis.
2. Move logic in `_post_extract` into a new package
3. Have an accounting_config for each study where the list of tests to run and parameters are defined
4. A template study generated by `kidsfirst new` will include the default set of accounting tests in the accounting_config


kf_lib_data_ingest/etl/accounting/test.py
```diff
@@ -0,0 +1,35 @@
+
```
znatty22: The module where all the accounting tests are defined and run.

kf_lib_data_ingest/etl/accounting/test.py
```diff
@@ -0,0 +1,35 @@
...
+    # return output, display_output
+    pass
+
+
+def a_has_b(df_dict, **inputs):
+    # return output, display_output
+    pass
+
+
+def run(accounting_config_path, extract_output):
```
znatty22: Load in the accounting config and run all the enabled tests that were listed

kf_lib_data_ingest/factory/templates/my_study/accounting_config.py
```diff
@@ -0,0 +1,43 @@
+"""
```
znatty22: An accounting config that looks something like this?

kf_lib_data_ingest/etl/accounting/test.py
```diff
@@ -0,0 +1,35 @@
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+def unique_concept_count(df_dict, **inputs):
```
znatty22: Each test returns some structured output and a string with the display output

kf_lib_data_ingest/etl/accounting/test.py
znatty22: This module defines all study wide accounting tests. These can be imported in the accounting configs

kf_lib_data_ingest/factory/templates/my_study/accounting_config.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # **** Auto-generated default tests **** #
+    {
+        'function': test.unique_concept_count,
+        'inputs': {
+            'concept': CONCEPT.PARTICIPANT.ID,
+            'expected': 100,  # we expect exactly 100 participants
+            'test': '=='
+        },
+        'enable': True
```
znatty22: Allows an ingest configurator to turn on/off certain tests. Could be good for debugging?

fiendish:
Something like this won't work without getting even more verbose in the configuration than this, and I already think that this is way too verbose on the configuration end. The tests need to be stage-specific, because some checks can only happen at certain times, and each stage has different output. I'm pretty sure that most of the configuration isn't needed and that test formats can be universally shared across all studies.

fiendish:
Any relationship tests, notably here a_has_b and family size, should only be done after transform. Otherwise you have to specifically reimplement transform in order to actually do those checks, which ignores why we're doing transform as a module in the first place.

All of the single value counting can (and should) be applied universally to every study with the only needed configuration to optionally specify how many you expect if you expect a certain number.

kf_lib_data_ingest/factory/templates/my_study/accounting_config.py
fiendish: Disagree. At least in this example. This is just a more verbose version of commenting out a single line that expresses the expected count for the given key.

znatty22:
Can we discuss the requirements of accounting (and anticipated needs) with stakeholders (@baileyckelly, @allisonheath)?

> Something like this won't work without getting even more verbose in the configuration than this, and I already think that this is way too verbose on the configuration end. The tests need to be stage-specific, because some checks can only happen at certain times, and each stage has different output. 

Totally agree that test's need to be stage-specific, but I don't think this should block the features below. The ingest configurator should only have to worry about knowing which accounting tests they want executed and any needed test parameters like expected counts, etc. They shouldn't need to worry about the fact that some tests get executed after extract and others happen after transform. The ingest library knows which standard tests to run and at what point in the pipeline. 

IMO, 

**Turning on/off tests**
Even though we will have tests that apply universally to all studies, it would still be useful for the ingest lib to allow someone to turn on/off tests on a per study basis. What if you don't have any genomic files yet but you still want to run accounting and validate participant, specimen counts and relationships? A study could still pass accounting and be approved for ingestion without the genomic files. Then later on when genomic files come in, we turn on the accounting tests for genomic files - both entity counts and relationship tests. Unless we think we'll never ingest a study without the genomic files, which I know I originally proposed but now I'm rethinking that. Or maybe we want to include accounting on the family entity for a particular study but not ALL studies.

**Modify test condition and parameters**
Also, what if I have a study where some participants have 1 sample that was taken and others have 2  samples that were taken. From a configuration standpoint, I want to be able to modify the relationship test to check that every participant has at least 1 specimen and less than or equal to 2 specimens: 
0 < participant.specimens.count.actual <= participant.specimens.count.expected.

**Custom tests**
We don't have to implement in this PR, but it would be really nice to easily define and plug in custom accounting tests. I think we'll definitely have some weird studies where the standard set of tests isn't enough. Maybe something like: I need to verify that this study has 40 trios and 20 duos? Maybe thats not the best example, but I *can* see this feature being very useful in the future. But maybe that's just me and I'm thinking too far ahead tho 🤷‍♀️

fiendish:
> What if you don't have any genomic files yet but you still want to run accounting and validate participant, specimen counts and relationships?

In my model all tests are treated as purely informative until directed otherwise. Failed tests won't stop anything from proceeding. They'll indicate a mismatch between discovery and expectation and move on. That is until you decide to set a flag saying that any mismatch should result in an abort, which should happen only after the study is perfect and passes all of the tests and is nominally ready for the portal in order to prevent regressions.

I see a goal mismatch between wanting tests to abort processing (which I think is implied in the model you propose) and also wanting to manually deactivate all of the tests that would abort said processing. In the model presented in #144 you get the benefit of the information that a test produces unexpected results while also letting you continue past those tests until you're ready to enforce strictness.

> Also, what if I have a study where some participants have 1 sample that was taken and others have 2 samples that were taken.

I'm convinced that all relationship tests have to happen after transform. This is just extract testing. I'm working on a simplified configuration for post-transform relationship tests that will cover some of this.

>  it would be really nice to easily define and plug in custom accounting tests

I agree. And I think it will be easier to defer that until after the standard post-transform tests are done.

fiendish:
> Can we discuss the requirements of accounting (and anticipated needs) with stakeholders 

Yes, but I don't think it should block 144 because nothing very interesting can safely happen after extract.

znatty22:
> With the strict_accounting toggle in #144 you get the benefit of the information when a test produces unexpected results while also letting you continue past those tests until you're ready to enforce strictness.

But what if I want to run ingest with strict accounting on for a study that only has participant and sample data available and no genomic files yet? I guess we **could** just set the expected count of genomic files to 0, effectively turning off that test. However, this approach might not be well suited to all types of test. I think it would still be nice to have a simple on/off switch. It would especially be useful in debugging when developing custom tests. 

Btw, its fine if this feature doesn't make it into this PR. I'm not advocating for that. But I am advocating for consideration of the feature in a future accounting PR because i think it would be very useful to users

Regarding your comments on post-transform 👍, I'll wait til post transform accounting tests to comment further

fiendish:
> what if I want to run ingest with strict accounting on for a study that ... no genomic files yet

The strict flag is for when you actually think that the data accurately represents the study and you want to prevent regressions from future changes. Otherwise you're lying to the test by saying "I expect there to be 100 genomic files here, lol jk fooled you I actually expected 0".

PR #146, opened 2019-02-25
♻️ refactor _source_file_to_df

fiendish:
read_table to read_csv because pandas has decided (sigh) to deprecate
the former

more help output

json file default loader

### DIFF
--- a/kf_lib_data_ingest/etl/extract/extract.py
+++ b/kf_lib_data_ingest/etl/extract/extract.py
@@ -42,6 +42,7 @@ class ExtractStage(IngestStage):
             self.extract_configs = [ExtractConfig(config_filepath)
                                     for config_filepath
                                     in extract_config_paths]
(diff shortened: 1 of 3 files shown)


kf_lib_data_ingest/etl/extract/extract.py
```diff
@@ -203,27 +204,44 @@ def _source_file_to_df(
...
-                df = pandas.read_table(f, sep=',', dtype=object, **load_args)
-            else:
-                raise ConfigValidationError(
-                    "Could not determine appropriate loader for data file",
-                    file_path
+            if file_path.endswith(('.xlsx', '.xls')):
+                load_func = read_excel_file
+            elif file_path.endswith(('.tsv', '.csv')):
+                load_args['sep'] = (
+                    load_args.pop('delimiter', None) or
```
znatty22: Nice. I always forget which one it is

Issue #147, opened 2019-02-25
🐛df[df.isna()] = None does not seem to work


znatty22:
Doesn't seem to work. I don't know why. But this works: `df = df.where((pandas.notnull(df)), None)`

Solution found here: https://stackoverflow.com/questions/14162723/replacing-pandas-or-numpy-nan-with-a-none-to-use-with-mysqldb

fiendish:
ugh. oops. I could have sworn...

PR #148, opened 2019-02-25
 🐛 extract data cleaning (bug #147)

fiendish:
Maybe we should just remove the line though? I didn't realize it before, but this replacement currently gets shadowed by the later call to intsafe_str anyway.

### DIFF
--- a/kf_lib_data_ingest/common/misc.py
+++ b/kf_lib_data_ingest/common/misc.py
@@ -5,7 +5,7 @@ import os
 import re
 from itertools import tee
 
-import numpy
+from pandas import isnull
 import yaml
 
 from kf_lib_data_ingest.common.type_safety import assert_safe_type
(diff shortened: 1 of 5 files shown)


znatty22:
Already talked about this on Slack, but reiterating here, if this goes in `intsafe_str`, lets call that function something else that's more descriptive of what its doing. I forget what `intsafe_str` does

kf_lib_data_ingest/etl/extract/extract.py
```diff
@@ -185,15 +189,31 @@ def _run(self):
...
+        # configurator to equate empty spreadsheet cells with empty strings.
+
+        # Typed loaders like pandas.read_json force us into storing numerically
+        # typed values, and then nulls, which read_json does not let you handle
+        # inline, cause pandas to convert perfectly good ints into ugly floats.
+        # So here we get any untidy values back to nice and tidy strings.
+
+        return df.applymap(
+            lambda x: to_str_with_floats_downcast_to_ints_first(
+                x, replace_na=True, na=''
```
znatty22: Why use empty strings here instead of None? If you leave the empty strings then transform will have to convert all empty strings to None later on anyway when its doing null processing.  Might as well do it here since, the output of _clean_up_df basically goes into transform

kf_lib_data_ingest/common/misc.py
```diff
@@ -48,25 +48,46 @@ def iterate_pairwise(iterable):
...
-def intsafe_str(val):
-    """Converts numbers to str while collapsing "1.0" to "1", etc.
-    in case of wrong numeric encoding of integers in a spreadsheet.
-
-    Args:
-        val: Any basic type
-
-    Returns:
-        string: Representation of `val`
+def to_str_with_floats_downcast_to_ints_first(val, replace_na=False, na=None):
```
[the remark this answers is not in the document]
fiendish: I wanted it clear that whole number floats are getting cast to ints :(

kf_lib_data_ingest/etl/extract/extract.py
fiendish: We just have to decide whether someone writing an extract configuration will look for "" or for None. Either way is ok.

znatty22: Yea None is better

kf_lib_data_ingest/common/misc.py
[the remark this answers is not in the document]
znatty22: Maybe just `numeric_to_str`? Idk..

PR #153, opened 2019-02-27
✨ Source ID to Target ID Translation

### LINKED ISSUE #151: TransformStage - Source ID to Target ID Translation
Happens on output of auto/guided transform. Translate source identifiers to target identifiers (i.e. Kids First IDs).

znatty22:
Will rebase after #174 

Closes #151


kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -112,6 +115,8 @@ def _run(self, target_api_config_path, auto_transform=False,
                 transform_fp = os.path.join(
                     self.ingest_config_dir, os.path.relpath(transform_fp))
 
+        id_cache_filepath = os.path.join(self.ingest_config_dir,
```
znatty22: I'm assuming the ID cache file will be stored in the ingest config dir. If this is a bad assumption then, we'll need to have `id_cache_filepath` as a parameter in `dataset_ingest_config.yml` and use the FileRetriever to fetch the file from wherever it is.

Figured reading it from ingest config dir for the first go around would be fine

fiendish: agree

kf_lib_data_ingest/config.py
```diff
@@ -19,3 +19,5 @@
 DEFAULT_LOG_OVERWRITE_OPT = True
 
 INGEST_PKG_TEMPLATE_NAME = 'my_study'
+
+DEFAULT_ID_CACHE_FILENAME = 'id_cache.json'
```
fiendish: Let's call all of these IDs UIDs or Service UIDs instead. We have a proliferation of ambiguously named ID fields

kf_lib_data_ingest/etl/transform/transform.py
```diff
@@ -232,6 +236,98 @@ def handle_nulls(self, target_instances, target_schema):
 
                     instance['properties'][attr] = mapped_value
 
+    def _source_to_target_ids(self, id_cache, target_instances):
```
fiendish: Please comment on what happens here if something can only be uniquely identified by a combination of multiple values.

kf_lib_data_ingest/etl/transform/transform.py
```diff
@@ -263,4 +359,12 @@ def _run(self, data_dict):
             self.logger.warning('Skipping null processing because no target '
                                 'schema was found')
 
+        # Source to target ID translation
+        id_cache = read_json(self.id_cache_filepath, default={})
+        if not id_cache:
+            self.logger.warning(
+                f'ID cache file not found {self.id_cache_filepath}')
```
fiendish: This message is a bit misleading. You will also get a Falsey response If your file exists but is empty.

kf_lib_data_ingest/config.py
znatty22: I know this is a long name, but its self-explanatory. What if I call this file `source_uid_to_target_uid_cache.json`?

Also, I made a separate issue to address the naming of the ID fields: https://github.com/kids-first/kf-lib-data-ingest/issues/178

kf_lib_data_ingest/etl/transform/transform.py
znatty22: Oo yes this is dealing with unique key to target id translation. I'll update the docstring and maybe even the method name

kf_lib_data_ingest/config.py
fiendish: uid_cache sounds good 👍

znatty22: Can we do `uid_cache.json` bc its a cache of target uids and also synthetic uids (currently known as unique keys)

fiendish: go for it

PR #163, opened 2019-02-28
 ✅ catch if load_func doesn't return a df

fiendish:


kf_lib_data_ingest/etl/extract/extract.py
```diff
@@ -167,12 +168,18 @@ def _run(self):
...
+                    data_path,
+                    do_after_load=extract_config.do_after_load,
+                    load_func=extract_config.load_func,
+                    **(extract_config.loading_params)
+                )
+            except ConfigValidationError as e:
+                raise type(e)(
+                    f'In extract config {extract_config.config_filepath}'
+                    f' : {str(e)}'
+                ).with_traceback(sys.exc_info()[2])
```
znatty22: Do you need `.with_traceback(sys.exc_info()[2])`? The log output doesn't seem to change whether you have it or not

fiendish: 🤷‍♂️ gone

Issue #183, opened 2019-03-12
Preserve original file's extension when fetching file


znatty22:
Currently the file retriever downloads the file's data to the temporary file object but does not preserve the original file extension. 

Everything has been working fine in extract stage because it uses the original `source_data_url` from the extract config to inspect file extension and choose a load_func, rather than the file extension of the fetched file. This works for everything except http type download URLs which don't contain the actual filename in the URL:
`http://localhost:8080/download/study/SD_00001111/file/SF_4H3NRQH4`

fiendish:
Question: Do we actually need to name the temp file, or is it sufficient to set a original_name member on the returned data object?

PR #184, opened 2019-03-12
♻️ Preserve original file's ext in fetched temp file

znatty22:
Addresses #183

### DIFF
--- a/kf_lib_data_ingest/common/file_retriever.py
+++ b/kf_lib_data_ingest/common/file_retriever.py
@@ -98,14 +98,12 @@ def _web_save(protocol, source_loc, dest_obj, auth=None, logger=None):
     :type protocol: str
     :param source_loc: address or path
     :type source_loc: str
-    :param dest_dir: Directory where temp file will be downloaded
-    :type dest_dir: str
-    :param delete: Flag indicating whether or not to delete file after download
-    :type delete: bool
+    :param dest_obj: Receives the data downloaded from the source file
+    :type dest_obj: File-like object
     :param auth: optional requests-compatible auth object
     :type auth: http://docs.python-requests.org/en/master/user/authentication/
 
-    :returns: the file-like object containing the downloaded data
+    :returns: None, the data goes to dest_obj
     """
     logger = logger or logging.getLogger(__name__)
     url = protocol + PROTOCOL_SEP + source_loc
--- a/kf_lib_data_ingest/common/file_retriever.py
+++ b/kf_lib_data_ingest/common/file_retriever.py
@@ -8,6 +8,7 @@ import logging
 import os
 import shutil
 import tempfile
+from urllib.parse import urlparse
 
 import boto3
 import botocore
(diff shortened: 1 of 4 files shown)


kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -182,18 +216,31 @@ def get(self, url, auth=None):
...
-                    protocol, path, self._files[url],
-                    auth=auth, logger=self.logger
-                )
+                # Settings for temporary file object that optionally deletes
+                # itself at the end of FileRetriever instance scope
+                # (if the enclosing folder isn't already being deleted)
+                delete = self.cleanup_at_exit and not self.__tmpdir
+
+                # Invoke http getter
+                # Note - Temp file must be created inside _web_save since we
```
znatty22: I tried to see if there was a way to set the NamedTemporaryFile's suffix after it's instantiated, but there doesn't seem to be a way unfortunately

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -182,18 +216,31 @@ def get(self, url, auth=None):
...
+                # won't know file ext until after file is requested
+                if (protocol == 'http') or (protocol == 'https'):
```
znatty22: I tried to see if there was a way to set the NamedTemporaryFile's suffix after it's instantiated, but there doesn't seem to be a way unfortunately

znatty22:
> Doing that wouldn't require any code outside of parsing the http response header.

Little confused by this. Regardless of what we do in terms of naming the temp file, we can't use the original source_data_url to get the file extension bc it may or may not have it. That's what extract stage was doing before: 
https://github.com/kids-first/kf-lib-data-ingest/blob/998f732048ce2c3816f3883e347cd7213d1c1ded/kf_lib_data_ingest/etl/extract/extract.py#L224
https://github.com/kids-first/kf-lib-data-ingest/blob/998f732048ce2c3816f3883e347cd7213d1c1ded/kf_lib_data_ingest/etl/extract/extract.py#L243

> Do we actually need to name the temp file, or is it sufficient to replace the name field or additionally set an original_name field on the returned data object and use that in extract?

I don't think we need to do anything differently? The tempfile coming from FileRetriever.get is assigned a random name (default behavior is random name has no file extension), and then the new code in this PR sets the extension to be the original file's extension using the `suffix` kwarg on NamedTemporaryFile

fiendish:
I propose https://github.com/kids-first/kf-lib-data-ingest/pull/185

The modification to _web_save is more compliant with RFC 6266 and the modification to get is much simpler.

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -88,23 +90,59 @@ def _web_save(protocol, source_loc, dest_obj, auth=None, logger=None):
...
     :param protocol: URL protocol identifier
     :type protocol: str
     :param source_loc: address or path
     :type source_loc: str
-    :param dest_obj: Receives the data downloaded from the source file
-    :type dest_obj: File-like object
+    :param dest_dir: Directory where temp file will be downloaded
+    :type dest_dir: str
+    :param delete: Flag indicating whether or not to delete file after download
+    :type delete: bool
```
fiendish: revert these param lines and the return line (though I'm not opposed to _also_ returning dest_obj. it just doesn't do that right now)

znatty22: None of the other getters return the dest obj so for now I'm going to revert back to what we had.

Issue #188, opened 2019-03-15
Ingest Package Testing/Data Validation


znatty22:
~**Note** - This functionality is already part of test runner in kf-ingest-packages repo. This issue is about moving it into the ingest library.  See [README section Test Your Package](https://github.com/kids-first/kf-ingest-packages/tree/add-readme) for background.~

**Motivation**
As ingest packages are being developed there should be an easy way to ensure: 

1. The ingest package will run through the ingestion pipeline without any exceptions or errors
2. Entity counts (participants, specimens, sequencing files) and relationship counts (a specimen's participant, a participant's specimens etc) match the expected values for the study
3. Additionally, as an ingest developer I might want to add other data validation tests for my package to check things not covered by 2 above. 

**Solution**
- Number 1 above will be addressed with #186 running ingest for a package in dry_run mode. If there are any exceptions the pipeline already exits with code 1
- Additionally, the data validation checks (2, 3 above) should always run when ingest runs. If any of the checks fail then the ingest app should exit with code 1.

znatty22:
Addresses https://github.com/kids-first/kf-ingest-packages/issues/17

znatty22:
Addressed by #202

PR #189, opened 2019-03-15
✨ Add Test CLI cmd

znatty22:
Addresses #188 and https://github.com/kids-first/kf-ingest-packages/issues/17

### DIFF
--- a/kf_lib_data_ingest/app/cli.py
+++ b/kf_lib_data_ingest/app/cli.py
@@ -1,10 +1,3 @@
-from kf_lib_data_ingest.app import settings
-from kf_lib_data_ingest.etl.ingest_pipeline import DataIngestPipeline
-from kf_lib_data_ingest.config import (
-    DEFAULT_TARGET_URL,
-    DEFAULT_LOG_LEVEL
-)
-from kf_lib_data_ingest.config import DEFAULT_LOG_LEVEL, DEFAULT_TARGET_URL
 """
 Entry point for the Kids First Data Ingest Client
 """
@@ -13,6 +6,13 @@ import logging
 import sys
 
 import click
+
+from kf_lib_data_ingest.config import (
+    DEFAULT_TARGET_URL,
+    DEFAULT_LOG_LEVEL
+)
+from kf_lib_data_ingest.app import settings
+from kf_lib_data_ingest.etl.ingest_pipeline import DataIngestPipeline
 
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 DEFAULT_LOG_LEVEL_NAME = logging._levelToName.get(DEFAULT_LOG_LEVEL)
@@ -54,13 +54,6 @@ def common_args_options(func):
               default=DEFAULT_TARGET_URL, show_default=True,
               help='Target service URL where data will be loaded into')(func)
 
-    func = click.option(
-        '--dry_run',
-        default=False,
-        is_flag=True,
-        help='A flag specifying whether to only pretend to send data to '
-        'the target service')(func)
-
     return func
 
 
@@ -76,6 +69,11 @@ def cli():
 
 
 @click.command()
+@click.option('--dry_run',
+              default=False,
+              is_flag=True,
+              help='A flag specifying whether to only pretend to send data to '
+              'the target service')
 @click.option('--use_async',
               default=False,
               is_flag=True,
@@ -89,7 +87,7 @@ def cli():
 #               'user guided transformation')
 @common_args_options
 def ingest(dataset_ingest_config_path, app_settings_filepath, log_level_name,
-           target_url, dry_run, use_async):
+           target_url, use_async, dry_run):
     """
     Run the Kids First data ingest pipeline.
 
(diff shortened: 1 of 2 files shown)


kf_lib_data_ingest/test_runner.py
```diff
@@ -0,0 +1,121 @@
...
+    with requests_mock.Mocker(real_http=True) as m:
+        msg = ('No POST or PATCH requests allowed to restricted '
+               f'target services in user defined unit tests! Check to make '
+               'sure your unit tests are not attempting to load data into '
+               f'target services at any of these URLs: {restricted_urls}')
+        for method_name in ['post', 'patch']:
+            for url in restricted_urls:
+                m.register_uri(method_name.upper(), url, exc=Exception(msg))
+
+        exit_code = pytest.main([user_defined_test_dir])
```
znatty22: pytest output currently goes to stdout. Should we try to add it to our ingest log?

fiendish: It could go in its own file

kf_lib_data_ingest/test_runner.py
```diff
@@ -0,0 +1,130 @@
...
+    :type kwargs: dict
+    """
+    logger.info('Running standard unit tests for '
+                f'{ingest_pipeline.ingest_config_dir} ...')
+    # TODO Enable this when dry run mode is implemented
+    # kwargs.update({
+    #     'dry_run': True
+    # })
+    ingest_pipeline.run(TARGET_SERVICE_CONFIG_PATH, **kwargs)
+    logger.info('✅  Standard unit tests passed!')
```
fiendish: What does passing mean here?

znatty22:
Your unit tests for your ingest package can be whatever you want them to be. If you wanted, you could read in the accounting outputs and write a couple tests to check for various things in there. 

> But like what, and how would it differ from accounting

It differs from accounting because accounting is a part of the ingest pipeline. `kidsfirst test` does two things. It runs ingest in dry run mode (standard test) to ensure your package runs without exceptions and then it runs any other use defined tests you have - which could check the accounting outputs IF you want. You don't have to implement any user tests tho if you don't want.

>, and why does it need a new CLI command

Later on when we implement an automated CI process for kf-ingest-packages, we will need a simple way to tell Jenkins, hey Jenkins run all unit tests for ingest package P. The Jenkins test script will do something like:
```
for package P in kf-ingest-packages/packages.yaml:
    `kidsfirst test <some package>`  
```

>  and a bunch of other methods?

What bunch of other methods?

fiendish:
> Your unit tests for your ingest package can be whatever you want them to be.

Can you give me a for instance?

> It differs from accounting because accounting is a part of the ingest pipeline.

That doesn't appear to differentiate them, because it runs ingest, and presumedly anything related to data that you want to be checking for is part of the accounting.

> What bunch of other methods?

The contents of test_runner.py and the contents of test_cli_test_cmd.py

znatty22:
I thought that accounting was going to be purely informative? So I was thinking if I wanted to have some validation on things that come out of accounting because I want to know right away that some checks failed, I could define those tests in my user defined unit tests. It wouldn’t repeat any accounting work, bc the tests could simply read in accounting output and perform tests on the content.

Yesterday when I was writing the transform module for test_study, I was thinking I wish I knew whether total # of participants, specimens, and files changed between my extract output and my transform module output so that I can tell if I screwed something up in my transform module. I would probably just have a test in my user tests that reads in post extract accounting output, reads in post transform accounting output and checks those #s. 

Also, there might be additional checks that someone wants to do that are not happening in the accounting checks. Maybe they want to verify # of females with particular sets of diagnoses, or verify that every family has one proband, or verify # of trios and duos, or # of bams vs crams vs other, etc.  

Perhaps if enough ingest package developers are writing the same unit tests, then this could be the criteria that determines whether or not the test gets added to the accounting library? Accounting could be like a library of checks some of which could become part of the standard test for an ingest package, and others of which could be optionally used in user defined unit tests.

In general, I can see this (user defined unit tests) being a very useful feature for ingest package developers because it gives us a lot of control over how we validate our ingest package code, and its very familiar and standardized because it uses pytest under the hood.

I'm going CC @parimalak and @liberaliscomputing bc I'd like to hear their opinions and see if they think it would be useful.

parimalak:
yeah I agree with @znatty22  to have specific study tests. I don't know how much development it required to get that feature in this sprint, but having tests could be useful for mentioned cases and also we could also can check if we find any discrepancy in future for particular sample or participant or genomic_file, it is easy to write a test and it also serves as documentation for that study for special cases we tested/checked and tests like this makes sure we are not overriding the changes we made.

liberaliscomputing:
Whenever I ingest a new study, I don't load transformed data right away but do sanity checking against  all entities' local caches (tsv files), running custom testing functions in notebook. Even though I've already pulled out **pseudo blessed numbers** while tinkering with data at a pre-extract stage, I still want to test out if it was yielded as will be rendered in the portal. So, I run custom tests after transformation and then load data in QA when the tests **seem** accounted (I said "seem" since still any ground truth about a study is not given to us). Since the ingest library and the data uploader are initially for internal use, it would be nicer if I can have room to write custom unit testing within the ingest ecosystem, not in a separate testing environment.

kf_lib_data_ingest/test_runner.py
liberaliscomputing: It's not very clear to me either. Maybe passing means the test runner didn't throw out en error?

fiendish:
I still don't see how what you all are talking about is any different from "I want to abort if accounting informs me that something has gone wrong" except now with a bunch of extra boilerplate code. So far I see some vague "I want to run very important tests" without saying why those very important tests shouldn't be part of the study's accounting.

> I was thinking I wish I knew whether total # of participants, specimens, and files changed between my extract output and my transform module output so that I can tell if I screwed something up in my transform module.

I think that sounds like a very good thing to check for during post-transform accounting, don't you?

Is the problem just that a stage's post-run accounting doesn't see that stage's input? That would be a much simpler change than this.

> there might be additional checks that someone wants to do that are not happening in the accounting checks

Why aren't they? Why do we need an entire new place to validate data when we're already adding a way to validate data?

znatty22:
I think we need to talk about this in person. 

I'm very confused bc it feels like we're advocating for the same thing? This PR wasn't initially targeted to achieve:

> Accounting was already planned to be specific to each study in addition to a shared set of standard tests. What makes these study-specific tests different from those study-specific tests?

But I actually think it does and in a pretty simple way.

To me, `simply exiting 1 instead of 0 if accounting says that something looks wrong`, is exactly what's implemented here

fiendish:
> To me, simply exiting 1 instead of 0 if accounting says that something looks wrong, is exactly what's implemented here

Then I'm very confused about what's going on in test_runner.py's `_run_standard_tests` and `_run_user_defined_tests`, because the very existence of those functions says otherwise to me.

PR #191, opened 2019-03-18
📝 Initial requirements doc for supporting auth

znatty22:


docs/design/auth.md
```diff
@@ -0,0 +1,124 @@
...
+source_data_url = 'http://api/download/study/SD_00001111/file/SF_NH4353C5'
+authentication = {
+    'type': OAuth2
+    'environment_vars': {
+        'provider': 'EGO_URL',
+        'client_id': 'CLIENT_ID_EGO',
+        'client_secret': 'CLIENT_SECRET_EGO'
+    }
+}
+```
```
fiendish: What...in every extract config? Yuck. I hate this.

dankolbman: Yeah, authentication shouldn't be configurable through ingest configs. It should be handled by whatever environment it is executing in.

Issue #196, opened 2019-03-22
add 500 to the requests retry status set


fiendish:
The dataservice throws status 500 when gen3 falls over. It should be added to the set of retry statuses.

PR #197, opened 2019-03-22
🐛 The dataservice throws 500 if gen3 does

fiendish:
Gen3 throws internal errors if you ask it to walk and chew gum at the same time. 
Then the dataservice forwards the 500 status, which basically kills GF endpoint access until https://github.com/kids-first/kf-api-dataservice/issues/487 gets fixed.


fiendish:
Fixes #196

fiendish:
The motivation was specific, but I do think that retrying a few times on a status 500 is probably the right thing to do anyway.

PR #198, opened 2019-03-25
♻️ move target uid lookup out of transform into load

fiendish:
This reduces a little bit of complexity by removing batch code that
can't solve the problem it needs to solve.

Load needs to know how to look up UIDs on the fly, because it will
regularly be the case that none or only some of them are prepopulated,
and if one wasn't prepopulated then everything dependent on it will need
to look it up to connect their foreign keys.

So if Load needs to know how to do it on the fly anyway, then there's no
value in doing it in bulk during transform.


fiendish:
WHAT DID YOU DO, LOL

PR #199, opened 2019-03-26
✨ Add auth to file retriever

znatty22:
First pass, adds support for fetching protected files. 

Supports:
- Basic auth
    - Was pretty much already implemented
- Token auth
    - Needed when developing and testing an ingest package locally
    - See https://github.com/kids-first/kf-api-study-creator/issues/137
- OAuth 2 auth
    - Needed when CI testing in a deployed environment
    - See https://github.com/kids-first/kf-api-study-creator/issues/110

**Notes**
- Things are where they are but they can always be moved around. Just a first pass to get basic auth logic and flow working
- This PR only deals with file retriever fetching protected files given auth parameters in an auth config dict. It does not deal with how that auth config dict is populated or how it gets fed into the ingest app. Will be in another PR


kf_lib_data_ingest/network/utils.py
```diff
@@ -0,0 +1,114 @@
+"""
```
znatty22: Should prob move all network related utils from common.misc into network.utils in another PR. 

Things to move
- requests_retry_session
- get_open_api_v2_schema

kf_lib_data_ingest/network/utils.py
```diff
@@ -0,0 +1,114 @@
...
+import cgi
+import logging
+
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+def get_file(url, dest_obj, **kwargs):
```
znatty22: This could be more generic - return either regular resource or file like obj but I left it file specific for now.

kf_lib_data_ingest/network/oauth2.py
```diff
@@ -0,0 +1,115 @@
...
+        logger.error(f'Unexpected response content from {oauth_token_url}, '
+                     f'status_code: {response.status_code}')
+        return
+
+    logger.info(f'Successfully fetched token,\n{pformat(resp_body)}')
+
+    return token
+
+
+def get_file(url, dest_obj, provider_domain=None, audience=None,
```
znatty22: This could prob be more generic - return either regular resource or file like obj but I left it file specific for now.

kf_lib_data_ingest/network/utils.py
```diff
@@ -0,0 +1,114 @@
...
+                            'filename.')
+
+        for chunk in response.iter_content(chunk_size=8192):
+            if chunk:
+                dest_obj.write(chunk)
+        dest_obj.seek(0)
+        response.close()
+
+    else:
+        logger.warning(f'Could not fetch {url}. Caused by '
```
znatty22: I'm wondering if `dest_obj` could be deleted here, but then we’d have to know if its a TemporaryFile or NamedTemporaryFile where delete=True. I didn't try very hard but on first glance it looks like `dest_obj` is just a file-like obj so it looks the same regardless of whether it was created via a TemporaryFile, NamedTemporaryFile or regular file obj

tests/conftest.py
```diff
@@ -25,6 +25,11 @@
                                      'transform_module.py')
 COMMAND_LINE_ERROR_CODE = 2
 
+TEST_AUTH0_DOMAIN = 'natashasingh.auth0.com'
```
znatty22: I think these will prob need to be replaced with kids first values later

tests/test_oauth2.py
```diff
@@ -0,0 +1,116 @@
...
+def test_get_service_token(info_caplog, provider_domain, audience, client_id,
+                           client_secret, expected_status, expected_log):
+    """
+    Test oauth2.get_service_token
+
+    1 - All valid inputs
+    2, 3 - Missing client_id and client_secret
+    4 - Wrong API Audience, Access denied
+    """
+    token = oauth2.get_service_token(provider_domain,
```
znatty22: Mock out the response don't call actual auth0 api

tests/test_oauth2.py
```diff
@@ -0,0 +1,116 @@
...
+        token_kwargs = {
+            'provider_domain': TEST_AUTH0_DOMAIN,
+            'audience': TEST_AUTH0_AUD,
+            'client_id': TEST_CLIENT_ID,
+            'client_secret': TEST_CLIENT_SECRET
+        }
+
+        fp = os.path.join(tmpdir, 'downloaded_file')
+        with open(fp, 'w+b') as dest_obj:
+            response = oauth2.get_file(url, dest_obj, **token_kwargs)
```
znatty22: Mock out token response don't actually call auth0 api

kf_lib_data_ingest/network/utils.py
```diff
@@ -0,0 +1,114 @@
...
+    environment variable where the parameter value is stored.
+
+    :param url: the URL to be inspected
+    :type url: str
+    :param auth_config: configuration dict of auth schemes and parameters
+    :type auth_config: dict
+
+    :returns selected_cfg: configuration dict of the selected auth scheme
+    """
+    selected_cfg = None
```
znatty22: Could use regex to match patterns too if we want, kept it simple for now

tests/test_file_retriever.py
```diff
@@ -30,7 +39,7 @@ def storage_dir(tmpdir_factory):
     return tmpdir_factory.mktemp('retrieved_files')
 
 
-@pytest.fixture(scope='session')
+@pytest.fixture(scope='function')
```
znatty22: Was interfering with other mocks bc `scope=session`

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -225,10 +287,20 @@ def get(self, url, auth=None):
...
+
+                # Set auth config if there is one and validate it
+                if self.auth_config:
+                    self._validate_auth_config(self.auth_config)
+
                 # Fetch the remote data
+                kwargs = {
+                    'auth': auth,
+                    'logger': self.logger,
+                    'auth_config': self.auth_config
```
znatty22: I wonder if we should keep `auth_config` specific to http type gets (and call it `http_auth_confg`) since we don't quite know how it will be used for the other types of gets. I could see it being used in the s3 get maybe to get AWS access credentials? Idk..

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -86,14 +101,30 @@ def _s3_save(protocol, source_loc, dest_obj, auth=None, logger=None):
     raise botocore.exceptions.NoCredentialsError()  # never got the file
 
 
-def _web_save(protocol, source_loc, dest_obj, auth=None, logger=None):
+def _web_save(protocol, source_loc, dest_obj, auth=None, auth_config=None,
```
fiendish: Why do we have two auth fields now? Can we unify them?

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -86,14 +101,30 @@ def _s3_save(protocol, source_loc, dest_obj, auth=None, logger=None):
...
     See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition # noqa E501
 
+    Authentication
+    --------------
+    Currently supports the following authentication schemes:
+        - HTTP Basic Authentication
+        - Token Authentication
+        - OAuth 2 Authentication
+
+    If `auth_config` is provided, inspect the URL to select the authentication
```
fiendish: This design seems very odd to me. If I'm asking for a specific URL, why would I pass in a bunch of different URLs and ask if the first one is among the bunch? I should already know the answer to that. If I know what I'm asking for, and I know what format I need to give you for an auth table, then I should also know how to find the right variables from in the table without passing the buck. I think either the caller should already know what values to use or the caller shouldn't be passing in an auth set at all and it should be loaded statically from a global configuration file _here_ (not at the caller). I think I prefer the latter.

kf_lib_data_ingest/common/file_retriever.py
znatty22: Yes, addressed 👍

kf_lib_data_ingest/common/file_retriever.py
znatty22: Addressed 👍

kf_lib_data_ingest/network/oauth2.py
```diff
@@ -0,0 +1,117 @@
...
+        'audience': audience,
+        'grant_type': 'client_credentials',
+        'client_id': client_id,
+        'client_secret': client_secret
+    }
+    oauth_token_url = f'https://{provider_domain}/oauth/token'
+    logging.info(f'Fetching token from {oauth_token_url} to access '
+                 f'{audience} resources')
+
+    response = requests.post(oauth_token_url, json=body)
```
fiendish: I think this should use the retry session

kf_lib_data_ingest/network/oauth2.py
```diff
@@ -0,0 +1,117 @@
...
+
+    :returns response: the requests.Response object
+    """
+    # Get access token to request resource
+    token = get_service_token(provider_domain,
+                              audience,
+                              client_id,
+                              client_secret)
+
+    # Something went wrong w getting token
```
fiendish: w?

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -239,3 +311,45 @@ def get(self, url, auth=None):
...
+        Validate config dict containing authentication parameters
+
+        `auth_configs` looks something like:
+
+            {
+                'http://api.com/files': {
+                    'type': 'basic',
+                    'username': 'the username',
+                    'password': 'the password'
+                    }
```
fiendish:
```suggestion
```

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -239,3 +311,45 @@ def get(self, url, auth=None):
...
+                'https://kf-api-study-creator.kids-first.io/download': {
+                    'type': 'oauth2',
+                    'client_id': 'STUDY_CREATOR_CLIENT_ID',
+                    'client_secret': 'STUDY_CREATOR_CLIENT_ID',
+                    'provider_domain': 'STUDY_CREATOR_AUTH0_DOMAIN',
+                    'audience': 'STUDY_CREATOR_API_IDENTIFIER',
+                },
+                's3://bucket/key': {
+                    'aws_profile': 'default'
+                    }
```
fiendish:
```suggestion
```

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -224,10 +280,26 @@ def get(self, url, auth=None):
                     dir=self.storage_dir,
                     delete=self.cleanup_at_exit and not self.__tmpdir
                 )
+
+                # Validate auth configs dict if it exists
+                # Select one auth config based inspection of url
+                auth_config = None
+                if self.auth_configs:
+                    self._validate_auth_configs(self.auth_configs)
```
fiendish: Following the above comment on preserving the ability to pass in an auth_config, you'd want to validate that too...
_validate_auth_configs could be changed to only validate a single auth config, validate the chosen auth config object down on line 291 under `if auth_config:` before the log statement, and then you could also loop the self.auth_configs items and validate its entries during init where it gets set.

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -162,6 +190,31 @@ def _file_save(protocol, source_loc, dest_obj, auth=None, logger=None):
...
+    :returns selected_cfg: configuration dict of the selected auth scheme
+    """
+    selected_cfg = None
+    if auth_configs:
+        for key, cfg in auth_configs.items():
+            if url.startswith(key):
+                selected_cfg = cfg
+                break
+    return selected_cfg
+
```
fiendish: I would like to advocate for removing the first and last two lines

```Python
    if auth_configs:
         for key, cfg in auth_configs.items():
             if url.startswith(key):
                 return cfg
```

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -224,10 +280,26 @@ def get(self, url, auth=None):
                     dir=self.storage_dir,
                     delete=self.cleanup_at_exit and not self.__tmpdir
                 )
+
+                # Validate auth configs dict if it exists
+                # Select one auth config based inspection of url
+                auth_config = None
```
fiendish: Instead of `auth_config = None` here, move it back to the method def as an optional input argument and check `if self.auth_configs and not auth_config` on the next line to preserve the ability to manually pass in a custom auth object for a request.

kf_lib_data_ingest/common/file_retriever.py
znatty22: 👍

kf_lib_data_ingest/common/file_retriever.py
znatty22: Ok, I removed that stuff bc I thought we didn't want that. I can put it back

kf_lib_data_ingest/common/file_retriever.py
znatty22: Ok, I removed that stuff bc I thought we didn't want that. I can put it back

kf_lib_data_ingest/network/utils.py
```diff
@@ -0,0 +1,63 @@
...
+"""
+import cgi
+import logging
+
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+def get_file(url, dest_obj, **kwargs):
```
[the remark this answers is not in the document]
znatty22: 👍

kf_lib_data_ingest/network/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    :type url: str
+    :param dest_obj: a file-like object that receives the data downloaded
+    :type dest_obj: a file-like object
+    :param kwargs: keyword args forwarded to requests.get
+    :type kwargs: dict
+    :returns response: requests.Response object
+    """
+
+    kwargs['stream'] = True
+    response = requests.get(url, **kwargs)
```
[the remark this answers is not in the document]
znatty22: 👍

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -196,23 +196,18 @@ def _select_auth_scheme(url, auth_configs):
...
-    See FileRetriever._validate_auth_config for details on auth_config format
-
     :param url: the URL to be inspected
     :type url: str
     :param auth_configs: optional dict mapping URL patterns to authentication
     schemes and necessary auth parameters
     :type auth_configs: dict
 
-    :returns selected_cfg: configuration dict of the selected auth scheme
+    :returns cfg: configuration dict of the selected auth scheme
```
fiendish: I don't think the variable name in :returns: docstring field is a real thing

kf_lib_data_ingest/network/oauth2.py
```diff
@@ -47,7 +47,8 @@ def get_service_token(provider_domain, audience, client_id, client_secret):
     logging.info(f'Fetching token from {oauth_token_url} to access '
                  f'{audience} resources')
 
-    response = requests.post(oauth_token_url, json=body)
+    response = requests_retry_session(connect=1).post(oauth_token_url,
```
fiendish: Is this another one of those mock problems?

znatty22: No I just didn't want it to be 10, I thought that was a lot? Idk what should it be

PR #210, opened 2019-04-01
✨ App settings

znatty22:
Adds ingest app level settings

App level settings include configuration that is likely to be common for a particular project using the ingest app and library. The app settings supply things like auth configuration, target API config, etc.


tests/data/test_study/extract_config_development.py
```diff
@@ -0,0 +1,92 @@
+from kf_lib_data_ingest.common import constants
```
znatty22: I wanted to reuse/import extract_configs/simple_tsv_example1.py_in each of my `extract_config<app_mode>.py` extract configs, but I couldn't do it because I need extract_configs to be a python package with a `__init__.py` file. Extract breaks when I add the `__init__.py` file since its not a real extract config. We should prob add a check to skip over this file

fiendish:
I'd like you to explain this one to me when you get a chance

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -76,6 +77,7 @@ def __init__(
         self.ingest_output_dir = os.path.join(self.ingest_config_dir, 'output')
 
         self.target_api_config_path = target_api_config_path
+        self.auth_configs = auth_configs
```
znatty22: Make sure to pop auth_configs from kwargs before logging ingest parameters. Could potentially have secrets and we don't want to log secrets

kf_lib_data_ingest/config.py
```diff
@@ -11,9 +11,7 @@
 TARGET_URL_KEY = 'target_url'
 DEFAULT_TARGET_URL = 'http://localhost:5000'
 KIDSFIRST_DATASERVICE_PROD_URL = 'http://kf-api-dataservice.kidsfirstdrc.org'
```
fiendish: It's weird that KIDSFIRST_DATASERVICE_PROD_URL isn't in the same place as TARGET_SERVICE_CONFIG_PATH

znatty22: 👍 Good point

znatty22: Actually, I think this is a left over thing from when things were a little different with null processing. Its not used anywhere in the source code anymore. The target_url is used when fetching target service schemas. 

KIDSFIRST_DATASERVICE_PROD_URL is only used in handle_nulls tests so I can move it to conftest. However, since we mock everything out, it probably doesn't even matter what URL we use for testing

PR #211, opened 2019-04-01
♻️ Separate stage tally format from checks

fiendish:
We can separate tallying from actual analysis and then do the analysis in a more controlled and comparable way.

Make a uniform stage tally format that each stage is repsonsible for building, and then we can give that to whatever tests we want.

### DIFF
--- a/kf_lib_data_ingest/cli.py
+++ b/kf_lib_data_ingest/cli.py
@@ -1,16 +1,14 @@
 """
 Entry point for the Kids First Data Ingest Client
 """
-import os
 import inspect
 import logging
+import os
+import sys
 
 import click
 
-from kf_lib_data_ingest.config import (
-    DEFAULT_TARGET_URL,
-    DEFAULT_LOG_LEVEL
-)
+from kf_lib_data_ingest.config import DEFAULT_LOG_LEVEL, DEFAULT_TARGET_URL
 
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 DEFAULT_LOG_LEVEL_NAME = logging._levelToName.get(DEFAULT_LOG_LEVEL)
@@ -74,9 +72,12 @@ def ingest(dataset_ingest_config_path, target_url, use_async, log_level_name):
         root_dir, 'target_apis', 'kids_first.py')
 
     # Run ingest
-    DataIngestPipeline(
+    err = DataIngestPipeline(
         dataset_ingest_config_path, target_api_config_path, **kwargs
     ).run()
+
+    if err:
+        sys.exit(1)
 
 
 @click.command(name='new')
--- a/kf_lib_data_ingest/cli.py
+++ b/kf_lib_data_ingest/cli.py
@@ -72,11 +72,11 @@ def ingest(dataset_ingest_config_path, target_url, use_async, log_level_name):
         root_dir, 'target_apis', 'kids_first.py')
 
     # Run ingest
-    err = DataIngestPipeline(
+    perfection = DataIngestPipeline(
         dataset_ingest_config_path, target_api_config_path, **kwargs
     ).run()
 
-    if err:
+    if not perfection:
         sys.exit(1)
 
 
--- a/kf_lib_data_ingest/common/misc.py
+++ b/kf_lib_data_ingest/common/misc.py
@@ -11,6 +11,7 @@ from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 from requests.exceptions import ConnectionError
 import yaml
+import jsonpickle
 
 from kf_lib_data_ingest.common.type_safety import assert_safe_type
 
@@ -42,15 +43,18 @@ def read_json(filepath, default=None):
     """
     if (default is not None) and (not os.path.isfile(filepath)):
         return default
-
     with open(filepath, 'r') as json_file:
-        return json.load(json_file)
+        json_str = json_file.read()
+    return jsonpickle.decode(json_str)
 
 
 def write_json(data, filepath):
+    json_str = json.loads(jsonpickle.encode(data))
     with open(filepath, 'w') as json_file:
-        json.dump(data, json_file, sort_keys=True, indent=4,
-                  separators=(',', ':'))
+        json.dump(
+            json_str,
+            json_file, sort_keys=True, indent=4, separators=(',', ':')
+        )
 
 
 def iterate_pairwise(iterable):
--- a/kf_lib_data_ingest/common/misc.py
+++ b/kf_lib_data_ingest/common/misc.py
@@ -39,8 +39,13 @@ def read_json(filepath, default=None, use_jsonpickle=True):
     does not exist, then return default.
 
     :param filepath: path to JSON file
-    :param default: default return value if file is not found
-    :param use_jsonpickle: whether to allow pickling of JSON-incompatible types
+    :type filepath: str
+    :param default: default return value if file not found, defaults to None
+    :type default: any, optional
+    :param use_jsonpickle: pickle JSON-incompatible types, defaults to True
+    :type use_jsonpickle: bool, optional
+    :return: your data
+    :rtype: dict
     """
     if (default is not None) and (not os.path.isfile(filepath)):
         return default
@@ -53,21 +58,26 @@ def read_json(filepath, default=None, use_jsonpickle=True):
             return json.load(json_file)
 
 
-def write_json(data, filepath, use_jsonpickle=True):
-    """
+def write_json(data, filepath, use_jsonpickle=True, **kwargs):
+    r"""
     Write Python dict to JSON file.
 
-    :param data: your python dict
-    :param filepath: path for new JSON file
-    :param use_jsonpickle: whether to allow pickling of JSON-incompatible types
+    :param data: your data
+    :type data: dict
+    :param filepath: where to write your JSON file
+    :type filepath: str
+    :param use_jsonpickle: pickle JSON-incompatible types, defaults to True
+    :type use_jsonpickle: bool, optional
+    :param \**kwargs: keyword arguments to pass to json.dump
     """
+    if 'indent' not in kwargs:
+        kwargs['indent'] = 4
+    if 'sort_keys' not in kwargs:
+        kwargs['sort_keys'] = True
     with open(filepath, 'w') as json_file:
         if use_jsonpickle:
             data = json.loads(jsonpickle.encode(data))
-        json.dump(
-            data,
-            json_file, sort_keys=True, indent=4, separators=(',', ':')
-        )
+        json.dump(data, json_file, **kwargs)
 
 
 def iterate_pairwise(iterable):
--- a/kf_lib_data_ingest/common/misc.py
+++ b/kf_lib_data_ingest/common/misc.py
@@ -33,26 +33,39 @@ def read_yaml(filepath):
         return yaml.load(yaml_file, Loader=yaml.FullLoader)
 
 
-def read_json(filepath, default=None):
+def read_json(filepath, default=None, use_jsonpickle=True):
     """
     Read JSON file into Python dict. If default is not None and the file
-    does not exist, then return default
+    does not exist, then return default.
 
     :param filepath: path to JSON file
     :param default: default return value if file is not found
+    :param use_jsonpickle: whether to allow pickling of JSON-incompatible types
     """
     if (default is not None) and (not os.path.isfile(filepath)):
         return default
+
     with open(filepath, 'r') as json_file:
-        json_str = json_file.read()
-    return jsonpickle.decode(json_str)
+        if use_jsonpickle:
+            json_str = json_file.read()
+            return jsonpickle.decode(json_str)
+        else:
+            return json.load(json_file)
 
 
-def write_json(data, filepath):
-    json_str = json.loads(jsonpickle.encode(data))
+def write_json(data, filepath, use_jsonpickle=True):
+    """
+    Write Python dict to JSON file.
+
+    :param data: your python dict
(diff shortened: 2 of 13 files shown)


kf_lib_data_ingest/common/stage.py
```diff
@@ -47,19 +47,32 @@ def write_output(self, output):
     def _run(self, *args, **kwargs):
         pass
 
-    def _postrun_analysis(self, run_output):
+    def _postrun_tally(self, run_output):
```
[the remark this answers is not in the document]
fiendish: How do you feel about `_postrun_discovery`

kf_lib_data_ingest/etl/stage_analyses.py
```diff
@@ -0,0 +1,83 @@
...
+import pandas
+from tabulate import tabulate
+
+from kf_lib_data_ingest.common.concept_schema import str_to_CONCEPT
+from kf_lib_data_ingest.etl.extract.extract import ExtractStage
+from kf_lib_data_ingest.etl.load.load import LoadStage
+from kf_lib_data_ingest.etl.transform.transform import TransformStage
+
+
+def check_counts(tally_sources, expected_counts):
```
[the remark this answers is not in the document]
fiendish: Some things are easy to configure by just writing a number. Some things require more.

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -151,20 +153,73 @@ def run(self):
...
+                self.check_stage_tally(stage_type, stage.logger, tally)
         except Exception as e:
             self.logger.exception(str(e))
             self.logger.info('Exiting.')
             exit(1)
 
         # Log the end of the run
         self.logger.info('END data ingestion')
+
+    def check_stage_tally(self, stage_type, logger, tally):
```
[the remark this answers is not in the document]
fiendish: √

kf_lib_data_ingest/common/stage.py
```diff
@@ -47,19 +47,32 @@ def write_output(self, output):
...
         """
-        Performs post-run accounting and other analysis. Here is where we'll
-        perform whichever automatic data consistency checks we can do
-        immediately after the current stage is done. Checks that can only be
-        comprehensively performed after later stages will not happen here.
+        Tallies counts in run output.
```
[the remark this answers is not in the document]
fiendish: √

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -151,20 +153,73 @@ def run(self):
...
         # Top level exception handler
         # Catch exception, log it to file and console, and exit
         try:
             # Iterate over stages and execute them
             output = None
-            for s in self._iterate_stages():
+            for stage in self._iterate_stages():
                 if not output:  # First stage gets no input
-                    output, checks_passed, accounting_data = s.run()
+                    output, tally = stage.run()
```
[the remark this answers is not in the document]
fiendish: √

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -151,20 +154,89 @@ def run(self):
...
         self.logger.info('END data ingestion')
+
+    def validate_stage_counts(
+        self, stage_type, logger, concept_discovery_dict
+    ):
+        """
+        Do some standard stage discovery tests
+        """
+
+        logger.info("Begin Basic Analysis")
```
znatty22:
```suggestion
        logger.info("Begin stage output data validation")
```

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -151,20 +154,89 @@ def run(self):
...
             exit(1)
 
         # Log the end of the run
         self.logger.info('END data ingestion')
+
+    def validate_stage_counts(
+        self, stage_type, logger, concept_discovery_dict
+    ):
+        """
+        Do some standard stage discovery tests
```
znatty22: More descriptive docstring plz

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -151,20 +154,89 @@ def run(self):
...
+        """
+
+        logger.info("Begin Basic Analysis")
+
+        passed_all = True
```
znatty22: This is a little hard to read. Something like this is more readable to me

```suggestion
        # Discovery data is missing
        if not concept_discovery_dict:
            logger.info('Discovery Data Not Found ❌')
            return False

        if not concept_discovery_dict.get('sources'):
            logger.info('Discovery Data Sources Not Found ❌')
            return False

        # Do stage counts validation
        passed_all = True

        passed, message = stage_analyses.check_counts(
            concept_discovery_dict.get('sources'),
            self.data_ingest_config.expected_counts
        )
        passed_all = passed_all and passed
        logger.info(message)

        # Compare stage counts to make sure we didn't lose concepts/concept attributes
        self.stage_discovery_sources[stage_type] = (
            concept_discovery_dict['sources']
        )
        if stage_type == TransformStage:
            if ExtractStage in self.stage_discovery_sources:
                passed, message = stage_analyses.compare_counts(
                    concept_discovery_dict.get('sources'),
                    self.stage_discovery_sources[ExtractStage]
                )
                passed_all = passed_all and passed
                logger.info(message)
            else:
                passed_all = False
                logger.info(
                    'No Extract Discovery Data Sources To Compare ❌'
                )

        logger.info("End Basic Analysis")
        return passed_all
```

kf_lib_data_ingest/etl/stage_analyses.py
```diff
@@ -0,0 +1,83 @@
+from pprint import pformat
```
znatty22: File still references `tally`

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -151,20 +154,89 @@ def run(self):
...
+                write_json(
+                    concept_discovery_dict,
+                    os.path.join(
+                        self.ingest_output_dir,
+                        stage_type.__name__ + '_concept_discovery.json'
+                    )
+                )
+
+                self.stage_outputs[stage_type] = output
+                self.validate_stage_counts(
```
znatty22: Shouldn't we collect the pass/fail result of each stage and then exit 1 if any are False?

kf_lib_data_ingest/etl/transform/transform.py
```diff
@@ -282,6 +282,13 @@ def _run(self, data_dict):
 
         return target_instances
 
+    def _postrun_discovery(self, run_output):
+        """
+        See the docstring for IngestStage._postrun_discovery.
+        """
+        # TODO: Implement after transform (or an intermediate stage) is fixed
+        # to produce output using CONCEPT keys, instead of target service keys.
+
```
znatty22:
```suggestion
return {'sources': None, 'links': None}
```

kf_lib_data_ingest/etl/ingest_pipeline.py
fiendish: √

kf_lib_data_ingest/etl/ingest_pipeline.py
fiendish: √

kf_lib_data_ingest/etl/stage_analyses.py
fiendish: √

kf_lib_data_ingest/etl/ingest_pipeline.py
fiendish: I forget why we want exiting?

kf_lib_data_ingest/etl/transform/transform.py
fiendish: √

kf_lib_data_ingest/etl/ingest_pipeline.py
znatty22: We talked about the following scenarios: 

Running `kidsfirst ingest <pkg>` does:
```
For each ingest stage
    - Run normal ingest pipeline stage
    - Compute standard counts using stage output
    - Validate standard counts against expectations
Then run user defined data validation tests if there are any
Collect pass/fail flags from all of the above. 
If any one of them are non-zero, exit 1. 
```
We *could* add a `--exit_on_validation_fail` flag as you had suggested before (you called it 
`--strict_accounting` I think.

Running `kidsfirst test <pkg>` effectively does:
```
kidsfirst ingest <pkg> --dry_run --exit_on_validation_fail (if we implement this opt)

Same steps as above except nothing is loaded into target service
```

fiendish: Right, ok.

> We could add a --exit_on_validation_fail flag as you had suggested before (you called it

Not going to though, because loading bad data into the dataservice isn't really a problem.

kf_lib_data_ingest/common/misc.py
```diff
@@ -42,15 +43,18 @@ def read_json(filepath, default=None):
     """
     if (default is not None) and (not os.path.isfile(filepath)):
         return default
-
     with open(filepath, 'r') as json_file:
```
znatty22: Instead of changing the default behavior of the `read_json` and `write_json`, can we add a keyword arg (maybe `jsonpickle=True/False` to both methods that will jsonpickle.decode/encode the dict?

I use these two methods from our lib outside of the ingest pipeline to read/write simple json that doesn't need pickling. Since this is a library as well as an app, other people might do the same so it would be good if we can accommodate

kf_lib_data_ingest/etl/extract/extract.py
```diff
@@ -388,86 +382,27 @@ def _run(self):
...
-                )
-            )
-
-        return all_checks_passed, counted, '\n'.join(message)
+                    sources[key][val].add(data_file)
+            for keyA in df.columns:
+                for keyB in df.columns:
+                    if keyB != keyA:
+                        for i in range(len(df)):
+                            links[keyA + keyB][df[keyA].iloc[i]].add(
```
znatty22: Can you add a delimiter to the link key? A key looks like this right now and its tough to read: 
```
"CONCEPT|BIOSPECIMEN|ALIQUOT_IDCONCEPT|BIOSPECIMEN|ANALYTE"
```

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -151,20 +155,98 @@ def run(self):
...
+                    stage.logger.info("End Basic Stage Output Validation")
         except Exception as e:
             self.logger.exception(str(e))
             self.logger.info('Exiting.')
-            exit(1)
+            sys.exit(1)
 
         # Log the end of the run
         self.logger.info('END data ingestion')
+        return not passed
```
znatty22: Heh? Shouldn't it be `return passed`?

kf_lib_data_ingest/common/misc.py
fiendish: > I use these two methods from our lib outside of the ingest pipeline to read/write simple json that doesn't need pickling. 

Are you saying that you actually want write_json to throw an exception on complex types? Because otherwise it shouldn't matter. If you don't include complex data, then nothing special will happen to your output.

kf_lib_data_ingest/cli.py
```diff
@@ -74,10 +72,13 @@ def ingest(dataset_ingest_config_path, target_url, use_async, log_level_name):
         root_dir, 'target_apis', 'kids_first.py')
 
     # Run ingest
-    DataIngestPipeline(
+    err = DataIngestPipeline(
         dataset_ingest_config_path, target_api_config_path, **kwargs
     ).run()
 
+    if err:
```
znatty22: Oh now I see why you return not passed in pipeline.run()... Hmm can we change it to, `no_errors` or `success` and then say if not `<success flag>` then exit 1?

kf_lib_data_ingest/common/misc.py
znatty22: Ohhh I see, I thought it would still write out the data in a verbose form with pickle stuff in there. I think this is fine then 👍

fiendish: eh, I'll throw in an argument anyway

kf_lib_data_ingest/etl/extract/extract.py
fiendish: √

kf_lib_data_ingest/etl/ingest_pipeline.py
fiendish: 🚀

kf_lib_data_ingest/cli.py
fiendish: √

kf_lib_data_ingest/common/misc.py
znatty22: If you're going to add the pick kwarg, then maybe also `**kwargs` to the method sigs of `read_json` and `write_json`  and then pass it into the `json.load/json.dump`. I prob should have added that anyway.

fiendish: hmm... I'll add it to dump but I think not to load. If you need special decoding, it seems like you should just be using the pickle option.

kf_lib_data_ingest/cli.py
```diff
@@ -74,10 +72,13 @@ def ingest(dataset_ingest_config_path, target_url, use_async, log_level_name):
         root_dir, 'target_apis', 'kids_first.py')
 
     # Run ingest
-    DataIngestPipeline(
+    perfection = DataIngestPipeline(
```
znatty22: lol

PR #214, opened 2019-04-04
🐳 Dockerize ingest app

### LINKED ISSUE #187: Dockerize Ingest App
Ingestion can be executed in a variety of ways using this library. For example, the CLI can be used to run the pipeline that comes with the library or someone could plug in parts of the library into their own ingestion pipeline (i.e. an Airflow DAG, several lambda functions, etc). 

In the near term, ingest jobs will run using the library's built in pipeline. It would be useful to dockerize the CLI based ingest app so that any machine or person wanting to run ingest jobs doesn't have to worry about getting python or setting up any other dependencies. 

To run ingest via the docker container, the machine or user would do something like:
```
docker exec kf-ingest-app (or whatever container name) ingest <package dir>

or 

docker exec kf-ingest-app (or whatever container name) test <package dir>
```
In the near term, the docker container will likely be used to run ingest dry-runs during the continuous integration process (testing) and actual ingest jobs in the continuous deployment process of https://github.com/kids-first/kf-ingest-packages. (More on that later)

znatty22:
Part of #217 and closes #187


tests/data/SD_ME0WME0W/dataset_ingest_config.yml
```diff
@@ -0,0 +1,46 @@
+
```
znatty22: SD_ME0WME0W study is used here to test the docker build during the build stage of CI [here](https://github.com/kids-first/kf-lib-data-ingest-config/blob/e7b893633ce3176260de7e6a27215caacc75e233/aws-ecs-service-type-1/ci-scripts/build_stage/build.sh#L22)

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+ENV KF_INGEST_APP_MODE production
+
+WORKDIR /app
+
+COPY setup.py requirements.txt /app/
+
+RUN pip install --upgrade pip && \
+    pip install --no-cache-dir -e .
+
+RUN apt-get update && apt-get install -y jq wget && \
```
fiendish: What if the system isn't apt-based?

tests/data/SD_ME0WME0W/dataset_ingest_config.yml
```diff
@@ -0,0 +1,46 @@
...
+    - 'study'
+    - 'study_file'
+    - 'family'
+    - 'participant'
+    - 'family_relationship'
+    - 'diagnosis'
+    - 'phenotype'
+    - 'outcome'
+    - 'biospecimen'
+    - read_group
```
fiendish: Why are some of these wrapped in ' and some not?

tests/data/SD_ME0WME0W/transform_module.py
```diff
@@ -0,0 +1,15 @@
...
+See documentation at
+https://kids-first.github.io/kf-lib-data-ingest/ for information on
+implementing transform_function.
+"""
+
+
+def transform_function(mapped_df_dict):
+    # Expected to return a single merged dataframe
+
+    return list(mapped_df_dict.values())[0]
```
fiendish: lol

Dockerfile
znatty22: Our docker image is built from the base python:3 image which is Debian based

tests/data/SD_ME0WME0W/dataset_ingest_config.yml
znatty22: Idk, I copied and pasted from test_study. I can fix 🛠

Dockerfile
fiendish: ⭕️

tests/data/SD_ME0WME0W/dataset_ingest_config.yml
```diff
@@ -16,9 +16,9 @@ target_service_entities:
     - read_group
```
fiendish: missed one

PR #219, opened 2019-04-08
✅ make test depend less on format

fiendish:


znatty22:
@fiendish I think we can implement these tests later, maybe after you finish the post-transform accounting?

PR #221, opened 2019-04-09
🐛 Fix rogue nulls

znatty22:
Fix unique key generation 
- Only build unique key strings if all required cols are present and the values are not null. 
- For optional cols that are null, use a value of Not Reported in the unique key string

Fix standard to target instance building
- Don't build target instances with a null id


znatty22:
> seems ok. it looks like you addressed it in two places, one downstream of the other. just in case?
Yes addressed two issues, but not just in case. Need both

PR #222, opened 2019-04-10
✨🐛 visibility attribute mirror

### LINKED ISSUE #218: visible==HIDDEN in kf_lib_data_ingest/target_apis/kids_first.py
The mapping for visible fields are all backwards. visible is _not_ hidden.

fiendish:
Get visibility right in transform.
If something is visible, then also record that it's not hidden, and vice versa.

fixes #218

### DIFF
--- a/kf_lib_data_ingest/etl/extract/extract.py
+++ b/kf_lib_data_ingest/etl/extract/extract.py
@@ -7,6 +7,7 @@ from pprint import pformat
 
 import pandas
 
+from kf_lib_data_ingest.common.concept_schema import concept_set
 from kf_lib_data_ingest.common.datafile_readers import read_excel_file
 from kf_lib_data_ingest.common.file_retriever import (
     PROTOCOL_SEP,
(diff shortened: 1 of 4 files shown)


kf_lib_data_ingest/etl/extract/extract.py
```diff
@@ -162,6 +163,27 @@ def _clean_up_df(self, df):
...
+            out_col[out_col.notnull()] = (
+                ~in_col[in_col.notnull()].astype(bool)
+            ).astype(object)
+            return out_col
+
+        for concept in concept_set:
+            if concept.HIDDEN in df and concept.VISIBLE not in df:
+                df[concept.VISIBLE] = flip_col(df[concept.HIDDEN])
+            elif concept.VISIBLE in df and concept.HIDDEN not in df:
+                df[concept.HIDDEN] = flip_col(df[concept.VISIBLE])
```
znatty22: Why do we want create the HIDDEN col if its not present? We only seem to be using VISIBLE in the kids_first.py target api config..

znatty22: Well I guess if we had both HIDDEN and VISIBLE then i see why, but see my comment above about mirroring..

fiendish: Because the target service mapping only allows for a direct map and while _ours_ uses visible, another might not.

PR #227, opened 2019-04-11
✨ Remove need to manually forward config elements

fiendish:
I don't think we should need to manually forward everything we want to use from a config file.

See e.g.

`self.target_concepts = self.contents.target_concepts` in TargetAPIConfig.__init__

### DIFF
--- a/kf_lib_data_ingest/etl/configuration/base_config.py
+++ b/kf_lib_data_ingest/etl/configuration/base_config.py
@@ -29,6 +29,10 @@ class AbstractConfig(ABC):
                 f'{self.config_filepath} failed validation'
             ) from e
 
+    def __getattr__(self, attr):
+        """ Forward attributes from self.contents """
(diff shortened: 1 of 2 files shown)


kf_lib_data_ingest/etl/configuration/base_config.py
```diff
@@ -29,6 +29,10 @@ def __init__(self, filepath, **kwargs):
                 f'{self.config_filepath} failed validation'
             ) from e
 
+    def __getattr__(self, attr):
```
znatty22: This method should prob only in PyModuleConfig bc it will fail for YamlConfig. YamlConfig's contents is a dict

znatty22: Or, you could check the type of contents and if its an attr, then call getattr, if its a dict, then call get?

Issue #27, opened 2018-08-03
Design, implement, test synchronous loader


znatty22:
The loader will receive a set of entity payloads that adhere to a standard schema defined by the target service. The load must be able to create or update entities in the target service. The load must save Kids First IDs upon creation of new entities so that later on, updates can be made. 

The loader is responsible for iterate over the payloads, translating any ID type fields into KF_IDs by looking up the ID in a cache of Kids First IDs, and then submitting the entity payload to the target service.

PR #229, opened 2019-04-11
✨ Load stage (part 1 of several)

### LINKED ISSUE #40: Determine ID caching strategy and format
KF IDs should probably be cached and persisted somewhere, maybe in a file which is persisted on S3.  Should there be one file per study or should all study's IDs be in a JSON file keyed by study and entity. Maybe a format could look like: 

```
study: {
    SD_00001111: {
        participant: {
            Schiffman-01: PT_00001111,
            ....
        } 
    }
}
```

### LINKED ISSUE #186: Add Dry Run Mode for Ingest Pipeline
@fiendish writing up your original idea in an issue ...

We should have a CLI option called something like `dry_run`. When supplied it will cause the ingest pipeline to execute all stages normally except for the load stage. During dry run mode the load stage would not actually send requests to the target service to modify data.

This mode of operation would be used to by:
- the ingest package developer while developing an ingest package for a dataset
- the CI system in https://github.com/kids-first/kf-ingest-packages to perform automated testing for ingest packages

**Load Stage Behavior:**

Note - UID cache (keeps a mapping of the unique id of the entity in the source data to the id assigned to entity by the target service) 

**Create new entity in target service**

Normal operation 
- Send POST with entity payload to the target service
- Update the UID cache with the new record of entity UID from source data, entity ID assigned by target service

Dry run operation
- Write the POST payload to the load stage output 
- Do not update UID cache

**Update existing entity in target service**

Normal operation
- Send PATCH with updated entity payload to the target service

Dry run operation
- Lookup entity's target service ID in UID cache
- Send GET to fetch entity from target service
- Update the entity payload from the GET response
- Save modified payload to load stage output

If someone looks at the load stage output from dry_run mode, they should see a preview of almost exactly what the data to be ingested would look like in the target service.

fiendish:
Part of #27 
closes #186
closes #40


kf_lib_data_ingest/app/cli.py
```diff
@@ -7,10 +7,7 @@
 
 import click
 
-from kf_lib_data_ingest.config import (
-    DEFAULT_TARGET_URL,
-    DEFAULT_LOG_LEVEL
-)
+from kf_lib_data_ingest.config import DEFAULT_LOG_LEVEL, DEFAULT_TARGET_URL
```
znatty22: I thought we agreed on isort multiline output mode of 3 - Vertical Hanging Indent?

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -6,28 +6,28 @@
 import logging
 import shutil
 import tempfile
-from urllib.parse import (
-    urlparse,
-    urlencode
-)
+from urllib.parse import urlencode, urlparse
```
znatty22: Same comment as above wrt isort

kf_lib_data_ingest/common/file_retriever.py
```diff
@@ -6,28 +6,28 @@
...
-    assert_safe_type,
-    assert_all_safe_type
-)
-from kf_lib_data_ingest.network import (
-    utils,
-    oauth2
+    assert_all_safe_type,
+    assert_safe_type
 )
+from kf_lib_data_ingest.network import oauth2, utils
```
znatty22: Same comment as above wrt isort

kf_lib_data_ingest/etl/load/load.py
```diff
@@ -1,80 +1,372 @@
...
 from urllib.parse import urlparse
 
+import pandas
+import requests
+from requests import RequestException
+from sqlite3worker import Sqlite3Worker, sqlite3worker
+
 from kf_lib_data_ingest.common.errors import InvalidIngestStageParameters
-from kf_lib_data_ingest.common.misc import read_json
+from kf_lib_data_ingest.common.misc import multisplit, requests_retry_session
```
znatty22: Same comment as above wrt isort

kf_lib_data_ingest/etl/load/load.py
```diff
@@ -1,80 +1,372 @@
...
+        :param entity_id: source unique ID for this entity
+        :type entity_id: str
+        :param entity_type: the name of this type of entity
+        :type entity_type: str
+        :param endpoint: which target service endpoint to hit
+        :type endpoint: str
+        :param body: map between entity keys and values
+        :type body: dict
+        :raises RequestException: Unhandled response error from the server
+        :return: The entity that the target service says was created
```
znatty22:
```suggestion
        :return: The entity that the target service says was created or updated
```

kf_lib_data_ingest/app/cli.py
fiendish: isort mode "...defines how from imports wrap **when they extend past the line_length limit**"

znatty22: Oh I see..I think we'll need to update the rest of the code base then at some point. I've been doing vertical hanging indent whenever I have more than 2 imports from a package

fiendish: lol

znatty22: I just like the way it looks 🤷‍♀️

PR #252, opened 2019-04-18
🐳 Docker container clean up after ingest ends

znatty22:
**Problem**
Docker container runs as root user or a user different than Jenkins user. When we run the container in the kf-ingest-packages CI pipeline on Jenkins host, we need to bind mount the kf-ingest-packages/kf_ingest_packages/packages directory into the container. When container is running ingest it writes logs/outputs as root into that bind mounted directory. When Jenkins tries to do its usual clean up it tries to delete the kf-ingest-packages/packages directory which contains logs/outputs that are owned by root.

**Solution**
Update docker container entrypoint so that it optionally cleans up (rm -rf) the ingest package dir after ingest finishes or fails. This is controlled by the CLEANUP env var which will contain the package name to delete. This is used in the CI script for kf-ingest-packages.

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -23,3 +23,13 @@ if [[ $KF_INGEST_APP_MODE = "production" ]]; then
 fi
 
 kidsfirst "$@"
+
+# During deployment - rm ingest package after ingest completes
+if [[ $CLEANUP ]]; then
+    PACKAGE_DIR="/data/packages/$CLEANUP"
+    if [[ -d $PACKAGE_DIR ]]; then
+        echo "Deleting ingest $PACKAGE_DIR ..."
+        rm -rf $PACKAGE_DIR
+        echo "Complete"
+    fi
+fi
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -30,15 +30,18 @@ if [[ $? -ne 0 ]]; then
     INGEST_ERROR=1
 fi
 
-# During deployment - rm ingest package after ingest completes
-if [[ $CLEANUP ]]; then
-    INGEST_PACKAGE_DIR="/data/packages/$CLEANUP"
+# If container was run by automated process - rm ingest package after ingest completes
+if [[ "$DELETE_INGEST_PKG" = true ]] && [[ "$INGEST_PKG_TO_DEL" ]]; then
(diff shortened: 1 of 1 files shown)


tests/data/SD_ME0WME0W/ingest_package_config.py
```diff
@@ -19,6 +19,11 @@
     'read_group_genomic_file'
 ]
 
+expected_counts = {
```
fiendish: unrelated

scripts/entrypoint.sh
```diff
@@ -22,4 +22,26 @@ if [[ $KF_INGEST_APP_MODE = "production" ]]; then
...
+    INGEST_ERROR=1
+fi
+
+# During deployment - rm ingest package after ingest completes
+if [[ $CLEANUP ]]; then
+    INGEST_PACKAGE_DIR="/data/packages/$CLEANUP"
+    if [[ -d $INGEST_PACKAGE_DIR ]]; then
+        echo "Start cleanup ..."
+        echo "Deleting ingest package $INGEST_PACKAGE_DIR"
+        rm -rf $INGEST_PACKAGE_DIR
```
fiendish: please quote this variable and put either `set -u` or `set -o nounset` at the start of the file

znatty22: k, will update

fiendish: Also, what if the script wasn't run by docker?

fiendish: Consider...

```
#!/bin/bash

mkdir -p "test"
CLEANUP="foo bar"
touch "test/$CLEANUP"
echo "CLEANING $CLEANUP"
ND="test/$CLEANUP"
rm $ND
echo "done"
```

prints:
> CLEANING foo bar
> rm: test/foo: No such file or directory
> rm: bar: No such file or directory
> done

vs 

```
#!/bin/bash

mkdir -p "test"
CLEANUP="foo bar"
touch "test/$CLEANUP"
echo "CLEANING $CLEANUP"
ND="test/$CLEANUP"
rm "$ND"
echo "done"
```

prints:
> CLEANING foo bar
> done

fiendish: I think that CLEANUP should be passed in as an argument to the script as possible and the check should not be based on whether the environment has something set.

PR #253, opened 2019-04-19
🔒 lock down merges

fiendish:
Until other people are more familiar with the project, I think one of us should be blockers on merging any PRs.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,6 @@
+*              @fiendish @znatty22 @allisonheath 
+/scripts/      @fiendish @znatty22 @allisonheath @alubneuski
+/.circleci/    @fiendish @znatty22 @allisonheath @alubneuski
+/.dockerignore @fiendish @znatty22 @allisonheath @alubneuski
+/Dockerfile    @fiendish @znatty22 @allisonheath @alubneuski
+/Jenkinsfile   @fiendish @znatty22 @allisonheath @alubneuski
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,6 @@
+*              @fiendish @znatty22
```
znatty22:
```suggestion
*              @fiendish @znatty22 @allisonheath 
```

CODEOWNERS
```diff
@@ -0,0 +1,6 @@
+*              @fiendish @znatty22 @allisonheath 
+/scripts/      @fiendish @znatty22 @allisonheath @alubneuski
```
znatty22: Is including `@fiendish @znatty22 @allisonheath` in lines after `*` repetitive? Since `*` says require review for everything in the repo?

znatty22: Also, I made a new team. You can replace all occurrences of`@fiendish @znatty22 @allisonheath` with `@kids-first/data-ingest`

fiendish: > Is including @fiendish @znatty22 @allisonheath in lines after * repetitive?

I don't think so, because the template for the owners file says that subsequent overlapping rules override prior settings.

fiendish: Though I guess if I switch the line order then maybe.

fiendish:
@znatty22 I also switched @alubneuski  to @kids-first/devops, so devops will need team write access to the repo too.

CODEOWNERS
znatty22: k 👍

Issue #232, opened 2019-04-12
extend expected_counts configuration to include values


fiendish:
Currently expected_counts includes something like e.g. `{'PARTICIPANT.ID': 5}`

We should extend it to do also something like `{'PARTICIPANT.SEX': {'Female': 5, 'Male': 4}}`

fiendish:
Details to consider: what does it mean to say that 5 instances of sex=Female were found during the extract stage if there are multiple files with overlapping sets of participant data? We probably need to link non-identifying keys to identifying keys. Like for instance, "there are 5 PARTICIPANT.ID where PARTICIPANT.SEX is Female in the same row."

znatty22:
Closing because this will be replaced with the new validation that will be implemented soon

PR #254, opened 2019-04-19
✨ Post transform accounting

### LINKED ISSUE #201: Post-Transform Accounting
extension of accounting that happens after the transform stage

fiendish:
closes #201

### DIFF
--- a/kf_lib_data_ingest/common/stage.py
+++ b/kf_lib_data_ingest/common/stage.py
@@ -182,12 +182,11 @@ class IngestStage(ABC):
                 if v is not None
             }
             # write to file
-            write_json(
-                self.concept_discovery_dict,
-                os.path.join(
+            out_file = os.path.join(
                     self.ingest_output_dir,
                     self.stage_type.__name__ + '_concept_discovery.json'
                 )
-            )
+            self.logger.debug(f"Writing discovered counts to {out_file}")
+            write_json(self.concept_discovery_dict, out_file)
 
         return output
--- a/kf_lib_data_ingest/common/stage.py
+++ b/kf_lib_data_ingest/common/stage.py
@@ -94,7 +94,7 @@ class IngestStage(ABC):
         containing them and 2) concept values map to lists of linked concept
         values
         """
-        return {'sources': None, 'links': None}
+        return {'sources': None, 'links': None, 'values': None}
 
     @abstractmethod
     def _validate_run_parameters(self, *args, **kwargs):
@@ -173,6 +173,15 @@ class IngestStage(ABC):
         self.concept_discovery_dict = self._postrun_concept_discovery(output)
         self.logger.info("Done counting what we got")
         if self.concept_discovery_dict:
+            # Undefault any defaultdicts. Defaultdicts are slightly dangerous
+            # to pass downstream to someone else's custom test code where they
+            # might accidentally add things while looking for keys.
+            self.concept_discovery_dict = {
+                k: dict(v)
+                for k, v in self.concept_discovery_dict.items()
+                if v is not None
+            }
+            # write to file
             write_json(
                 self.concept_discovery_dict,
                 os.path.join(
--- a/kf_lib_data_ingest/etl/ingest_pipeline.py
+++ b/kf_lib_data_ingest/etl/ingest_pipeline.py
@@ -242,12 +242,12 @@ class DataIngestPipeline(object):
         also whether any values were lost between Extract and Transform.
         """
         stage_name = stage.stage_type.__name__
-        stage.logger.info(f'Begin Basic Stage Output Validation')
+        stage.logger.info('Begin Basic Stage Output Validation')
         discovery_sources = stage.concept_discovery_dict.get('sources')
 
         # Missing data
         if not discovery_sources:
-            stage.logger.info(f'Discovery Data Sources Not Found ❌')
+            stage.logger.info('❌ Discovery Data Sources Not Found')
             return False
 
         passed_all = True
@@ -281,7 +281,7 @@ class DataIngestPipeline(object):
                 # Missing data
                 passed_all = False
                 stage.logger.info(
-                    'No ExtractStage Discovery Data Sources To Compare ❌'
+                    '❌ No ExtractStage Discovery Data Sources To Compare'
                 )
         stage.logger.info('End Basic Stage Output Validation')
 
--- a/kf_lib_data_ingest/etl/ingest_pipeline.py
+++ b/kf_lib_data_ingest/etl/ingest_pipeline.py
@@ -213,28 +213,28 @@ class DataIngestPipeline(object):
         passed_all = True
 
         # Do stage counts validation
-        passed, message = stage_analyses.check_counts(
-            discovery_sources, self.data_ingest_config.expected_counts
+        passed = stage_analyses.check_counts(
+            discovery_sources, self.data_ingest_config.expected_counts,
+            stage.logger
         )
         passed_all = passed_all and passed
-        stage.logger.info(message)
 
         # Compare stage counts to make sure we didn't lose values between
         # Extract and Transform
         if stage.stage_type == TransformStage:
             extract_disc = self.stages[ExtractStage].concept_discovery_dict
             if extract_disc and extract_disc.get('sources'):
-                passed, message = stage_analyses.compare_counts(
+                passed = stage_analyses.compare_counts(
                     ExtractStage.__name__,
                     extract_disc['sources'],
                     TransformStage.__name__,
                     {
                         k: v for k, v in discovery_sources.items()
                         if UNIQUE_ID_ATTR not in k
-                    }
+                    },
+                    self.logger
                 )
                 passed_all = passed_all and passed
-                stage.logger.info(message)
             else:
                 # Missing data
                 passed_all = False
--- a/kf_lib_data_ingest/etl/ingest_pipeline.py
+++ b/kf_lib_data_ingest/etl/ingest_pipeline.py
@@ -18,6 +18,7 @@ from kf_lib_data_ingest.etl.load.load import LoadStage
 from kf_lib_data_ingest.etl.transform.auto import AutoTransformStage
 from kf_lib_data_ingest.etl.transform.guided import GuidedTransformStage
 from kf_lib_data_ingest.etl.transform.transform import TransformStage
+from kf_lib_data_ingest.common.concept_schema import UNIQUE_ID_ATTR
 
 
 class DataIngestPipeline(object):
@@ -224,8 +225,13 @@ class DataIngestPipeline(object):
             extract_disc = self.stages[ExtractStage].concept_discovery_dict
             if extract_disc and extract_disc.get('sources'):
                 passed, message = stage_analyses.compare_counts(
-                    discovery_sources,
-                    extract_disc.get('sources')
+                    ExtractStage.__name__,
+                    extract_disc['sources'],
+                    TransformStage.__name__,
+                    {
+                        k: v for k, v in discovery_sources.items()
+                        if UNIQUE_ID_ATTR not in k
+                    }
                 )
                 passed_all = passed_all and passed
                 stage.logger.info(message)
--- a/kf_lib_data_ingest/etl/ingest_pipeline.py
+++ b/kf_lib_data_ingest/etl/ingest_pipeline.py
@@ -1,3 +1,4 @@
+import contextlib
 import inspect
 import logging
 import os
@@ -7,6 +8,8 @@ from pprint import pformat
 import pytest
 
 import kf_lib_data_ingest.etl.stage_analyses as stage_analyses
+from kf_lib_data_ingest.common.concept_schema import UNIQUE_ID_ATTR
+from kf_lib_data_ingest.common.misc import timestamp
 from kf_lib_data_ingest.common.type_safety import assert_safe_type
 from kf_lib_data_ingest.config import DEFAULT_TARGET_URL
 from kf_lib_data_ingest.etl.configuration.ingest_package_config import (
@@ -18,7 +21,6 @@ from kf_lib_data_ingest.etl.load.load import LoadStage
 from kf_lib_data_ingest.etl.transform.auto import AutoTransformStage
 from kf_lib_data_ingest.etl.transform.guided import GuidedTransformStage
 from kf_lib_data_ingest.etl.transform.transform import TransformStage
-from kf_lib_data_ingest.common.concept_schema import UNIQUE_ID_ATTR
 
 
 class DataIngestPipeline(object):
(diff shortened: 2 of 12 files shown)


fiendish:
@znatty22 Transform stage appears to be inserting nans?

See eg
```
Column CONCEPT|BIOSPECIMEN|ALIQUOT_ID not equal
ExtractStage {'4', '6', '28', '32', '48', '2', '', '11', '5', '30', '3', '47', '9', '58', '54', '49', '63', '52', '17', '38', '35', '21', '46', '15', '26', '45', '40', '56', '19', '18', '43', '13', '31', '12', '10', '24', '60', '39', '59', '16', '33', '8', '37', '20', '53', '23', '7', '29', '61', '57', '22', '42', '55', '1', '51', '62', '41', '14', '44', '50', '36', '25', '34', '27'}
vs
TransformStage {'4', '6', '', nan, '2', '28', '32', '48', '11', '5', '30', '3', '47', '9', '58', '54', '49', '63', '52', '17', '38', '35', '21', '46', '15', '26', '45', '40', '56', '19', '18', '43', '13', '31', '12', '10', '24', '60', '39', '59', '16', '33', '8', '37', '20', '53', '23', '7', '29', '61', '57', '22', '42', '1', '55', '51', '62', '41', '14', '44', '50', '36', '25', '34', '27'}
Difference is:
{nan}

Column CONCEPT|BIOSPECIMEN|ANALYTE not equal
ExtractStage {'DNA'}
vs
TransformStage {nan, 'DNA'}
Difference is:
{nan}

Column CONCEPT|BIOSPECIMEN|ID not equal
ExtractStage {'4', '6', '28', '32', '48', '11', '5', '30', '47', '9', '58', '54', '49', '63', '52', '17', '38', '35', '21', '46', '15', '26', '45', '40', '56', '19', '18', '43', '13', '31', '12', '10', '24', '60', '39', '59', '16', '33', '8', '37', '20', '53', '23', '7', '29', '61', '57', '22', '42', '55', '51', '62', '41', '14', '44', '50', '36', '25', '34', '27'}
vs
TransformStage {'4', nan, '6', '28', '32', '48', '11', '5', '30', '47', '9', '58', '54', '49', '63', '52', '17', '38', '35', '21', '46', '15', '26', '45', '40', '56', '19', '18', '43', '13', '31', '12', '10', '24', '60', '39', '59', '16', '33', '8', '37', '20', '53', '23', '7', '29', '61', '57', '22', '42', '55', '51', '62', '41', '14', '44', '50', '36', '25', '34', '27'}
Difference is:
{nan}
```

kf_lib_data_ingest/etl/stage_analyses.py
```diff
@@ -80,5 +80,35 @@ def check_counts(discovery_sources, expected_counts):
...
+    setB = set(discovery_sources_two.keys())
+    if not (setA == setB):
+        message.append('')
+        message.append(f'Column keys not equal')
+        message.append(f'{name_one} {setA}')
+        message.append('vs')
+        message.append(f'{name_two} {setB}')
+        message.append('Difference is:')
+        message.append(setA ^ setB)
+        passed = False
```
znatty22: Do you think this should always result in an error? Will there ever be a scenario where someone chooses not to merge all of their extract dfs into the single df?

kf_lib_data_ingest/etl/stage_analyses.py
```diff
@@ -80,5 +80,35 @@ def check_counts(discovery_sources, expected_counts):
     return passed, '\n'.join(message)
 
 
-def compare_counts(discovery_sources_one, discovery_sources_two):
-    False, "Compare Counts NYI"
+def compare_counts(
```
znatty22: Why not:
```suggestion
def compare_counts(
    name_one, discovery_sources_one, name_two, discovery_sources_two
):
    message = [f'COMPARING COUNTS BETWEEN {name_one} and {name_two}']
    passed = True

    setA = set(discovery_sources_one.keys())
    setB = set(discovery_sources_two.keys())
    diff = setA ^ setB
    if diff:
        message.append('')
        message.append(f'Column keys not equal')
        message.append(f'{name_one} {setA}')
        message.append('vs')
        message.append(f'{name_two} {setB}')
        message.append('Difference is:')
        message.append(diff)
        passed = False

    for k, v in discovery_sources_one.items():
        setA = set(v.keys())
        setB = set(discovery_sources_two[k].keys())
        diff = setA ^ setB
        if diff:
            message.append('')
            message.append(f'Column {k} not equal')
            message.append(f'{name_one} {setA}')
            message.append('vs')
            message.append(f'{name_two} {setB}')
            message.append('Difference is:')
            message.append(diff)
            passed = False

    return passed, '\n'.join([str(m) for m in message])
```

kf_lib_data_ingest/etl/transform/guided.py
```diff
@@ -198,3 +205,44 @@ def _do_transform(self, data_dict):
...
+            for keyA in df.columns:
+                vA = row[keyA]
+                for keyB in df.columns:
+                    if keyB != keyA:
+                        vB = row[keyB]
+                        if vA and vB:
+                            # links entry
+                            links[keyA + '::' + keyB][vA].add(vB)
+
+                            # values entry
```
znatty22: Maybe best to explain in person, but I'm not sure I understand what this is doing? Maybe for relationship counting?

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -224,8 +225,13 @@ def check_stage_counts(self, stage):
...
+                    ExtractStage.__name__,
+                    extract_disc['sources'],
+                    TransformStage.__name__,
+                    {
+                        k: v for k, v in discovery_sources.items()
+                        if UNIQUE_ID_ATTR not in k
+                    }
                 )
                 passed_all = passed_all and passed
                 stage.logger.info(message)
```
znatty22: I think the compare_counts output is too much for an info log level. This type of detailed output should probably be logged at the debug level, and only summaries should be logged at info level. So log this at debug level: 
```
Column CONCEPT|PARTICIPANT|CONSENT_TYPE not equal
ExtractStage {'GRU', 'HMB-IRB', 'DS-OC-PUB-MDS'}
vs
TransformStage {'GRU', None, 'HMB-IRB', 'DS-OC-PUB-MDS'}
Difference is:
{None}
```
And maybe this at info level: 
```
❌ Column values for CONCEPT|PARTICIPANT|CONSENT_TYPE not equal between ExtractStage and TransformStage
# of different values = 1
```

kf_lib_data_ingest/common/stage.py
```diff
@@ -173,12 +173,18 @@ def run(self, *args, **kwargs):
         self.concept_discovery_dict = self._postrun_concept_discovery(output)
         self.logger.info("Done counting what we got")
         if self.concept_discovery_dict:
-            write_json(
-                self.concept_discovery_dict,
-                os.path.join(
+            # undefault any defaultdicts
```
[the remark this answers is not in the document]
fiendish: defaultdicts are slightly dangerous to pass downstream to someone else's custom test code where they might accidentally add things while inspecting

kf_lib_data_ingest/etl/stage_analyses.py
fiendish: I don't know.

kf_lib_data_ingest/etl/ingest_pipeline.py
fiendish: √

kf_lib_data_ingest/etl/stage_analyses.py
znatty22: Well the other option is just that this doesn't result in a fail. Idk, I could go either way here

kf_lib_data_ingest/etl/stage_analyses.py
fiendish: potato potato

kf_lib_data_ingest/etl/stage_analyses.py
fiendish: I'd rather set this to fail now and try to find alternate ways of dealing with your proposed scenario.

kf_lib_data_ingest/etl/stage_analyses.py
fiendish: √

kf_lib_data_ingest/etl/transform/guided.py
fiendish: in preparation for addressing #232

kf_lib_data_ingest/etl/stage_analyses.py
znatty22: 👍

kf_lib_data_ingest/etl/ingest_pipeline.py
fiendish: I'm not sure about this. The info messages become not useful anymore by doing that.

fiendish:
> I don't see any relationship count tests, is that coming in another PR?

Yeah

> In addition to logging results of counting tests in line, should we create a single page count report sectioned by stage and write that to file? Then instead of saying this at the end of the log:

Hmm. Could do. It would require another logger.

fiendish:
How about this for your second log file idea? https://github.com/kids-first/kf-lib-data-ingest/pull/254/commits/c7d80f28b5247c9a20c3201517489c095e604990

kf_lib_data_ingest/etl/stage_analyses.py
```diff
@@ -70,15 +65,57 @@ def check_counts(discovery_sources, expected_counts):
...
+):
+    logger.info(f'COMPARING COUNTS')
+    passed = True
+
+    setA = set(discovery_sources_one.keys())
+    setB = set(discovery_sources_two.keys())
+    diff = setA ^ setB
+    if diff:
+        logger.info(
+            f'❌ Column keys not equal between {name_one} and '
```
znatty22: So this is weird...in my text editor and log its showing there is no space between ❌and Column, however in this diff preview there is actually a space. 

Here's what my log looks like
image

znatty22: Also can `Column keys` -> `Column names`?

kf_lib_data_ingest/etl/stage_analyses.py
```diff
@@ -70,15 +65,57 @@ def check_counts(discovery_sources, expected_counts):
...
+            str(diff)
+        )
+        passed = False
+
+    for k, v in discovery_sources_one.items():
+        setA = set(v.keys())
+        setB = set(discovery_sources_two[k].keys())
+        diff = setA ^ setB
+        if diff:
+            logger.info(
```
znatty22: Seems redundant with line 111

```suggestion
            logger.info(f'❌ Column values for {k} not equal between {name_one} and '
                f'{name_two}  ( # of different values = {len(diff)})'
```
I like keeping the error to a single log statement instead of two separate ones. I think it's easier to search/text process

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -212,29 +217,44 @@ def check_stage_counts(self, stage):
         passed_all = True
 
         # Do stage counts validation
-        passed, message = stage_analyses.check_counts(
-            discovery_sources, self.data_ingest_config.expected_counts
+
+        # Split log messages to another file just for this step
```
znatty22: This doesn't really give us what we want. Right now the `counts_for_ingest.log` seems to be overwritten each time `check_stage_counts` is called and what we end up with is just the count logs for the GuidedTransformStage since that is the last stage for which count analysis actually runs. Additionally, it looks like only INFO level logs get output to this file.

Hmm I don't know if using a separate logger is the way to go here. I'm sorry if my comment was misleading when I gave the example with a `.log` ext. I think what we want is a count analysis test report which contains all of the detailed results of the count analysis for each stage (and perhaps even a summary at the top?). Not sure of the best way to accomplish that. As for the logs, I think what you're currently logging at INFO level is fine. And then maybe we just don't log anything additional at the DEBUG log level. Instead, that detailed info can be found in the count analysis test report.

My intention with this was just to avoid cluttering the ingest log with things that may be better in a separate report, and also make it easy for the user to see exactly what went wrong with count analysis by directing them to a report that *only* contains count analysis results.

fiendish:
the test will start passing after #273 merges

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -180,10 +189,14 @@ def run(self):
...
-                    passed = self.check_stage_counts(stage) and passed
+                    passed, messages = self.check_stage_counts(stage)
+                    all_passed = all_passed and passed
+                    all_messages.extend(messages)
+
+            self._log_count_results(all_passed, all_messages)
 
             # Run user defined data validation tests
-            passed = self.user_defined_tests() and passed
+            all_passed = all_passed and self.user_defined_tests()
```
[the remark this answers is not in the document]
fiendish: oops, you're right. dratted short circuiting.

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -96,6 +99,11 @@ def __init__(
         self.log_file_path = setup_logger(log_dir, **log_kwargs)
         self.logger = logging.getLogger(type(self).__name__)
 
+        head, tail = os.path.split(self.log_file_path)
+        self.counts_file_path = os.path.join(head, 'counts_for_' + tail)
+        with contextlib.suppress(FileNotFoundError):
```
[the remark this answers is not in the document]
fiendish: > Why not use try/except?

🤷‍♂ 2 lines (import+suppress) vs 3 (try+except+pass)? tomato tomato

> shouldn't we clear all previous stage output rather than just the count log..why did we decide to do it for just count logs now

The counts file was originally written piecemeal in append mode after each stage. That's no longer done, so I'll just remove this.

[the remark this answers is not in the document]
znatty22: > All of the files naturally replace themselves

Only if the current run is overwriting existing files. What if someone removed an extract config from their package and then re-ran? The extract output for the removed extract config would still be there bc there is nothing clearing out the entire output folder. I think we should be doing this so that each run has a fresh set of output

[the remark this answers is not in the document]
znatty22: Probably should do that ^ in another PR tho

fiendish:
This works with the current state of #285

kf_lib_data_ingest/etl/stage_analyses.py
```diff
@@ -57,28 +47,92 @@ def check_counts(discovery_sources, expected_counts):
...
+            )
+            messages.extend(_format(msg, comparison_df))
+            passed = False
+
+    return passed, messages
+
+
+def _compare(name_one, list_one, name_two, list_two):
+    indicator = 'Errors'
+    one = pandas.DataFrame({name_one: list(list_one)}).dropna()
```
znatty22: You can filter nulls while constructing the list so that you don't have to iterate over the same values twice. See #276

fiendish: more verbose but ok

kf_lib_data_ingest/etl/stage_analyses.py
```diff
@@ -57,28 +49,113 @@ def check_counts(discovery_sources, expected_counts):
...
+        messages.extend(_format(msg, comparison_df))
+        passed = False
+    else:
+        msg = f'✅ Column names are equal between {name_one} and {name_two}'
+        messages.append(msg)
+        messages.append(SEP)
+
+    good_msg_ag = []
+    bad_msg_ag = []
+    for k in (one_keys | two_keys):
```
znatty22: Should this be 
```suggestion
    for k in set(one_keys | two_keys):
```

fiendish: not necessary. set operations work on dict_keys

PR #273, opened 2019-04-29
🐛 click.Choice needs a list, not an iterator

### LINKED ISSUE #272: cli log_level choices fail to present if run multiple times inside the same test environment
the second programmatic invocation of a cli function will exit 2 when given log_level because no matter what you give it will say that it's not one of the possible options

fiendish:
the click.Choice docstring says:
"You should only pass a list or tuple of choices. Other iterables
    (like generators) may lead to surprising results."

fixes #272

### DIFF
--- a/kf_lib_data_ingest/app/cli.py
+++ b/kf_lib_data_ingest/app/cli.py
@@ -42,8 +42,9 @@ def common_args_options(func):
         'supplied via config, the default log_level for the ingest lib will '
         f'be used: {DEFAULT_LOG_LEVEL_NAME}')
     func = click.option('--log_level', 'log_level_name',
-                        type=click.Choice(map(str.lower,
-                                              logging._nameToLevel.keys())),
+                        type=click.Choice(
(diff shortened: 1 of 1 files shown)


kf_lib_data_ingest/app/cli.py
```diff
@@ -42,8 +42,9 @@ def common_args_options(func):
         'supplied via config, the default log_level for the ingest lib will '
         f'be used: {DEFAULT_LOG_LEVEL_NAME}')
     func = click.option('--log_level', 'log_level_name',
-                        type=click.Choice(map(str.lower,
-                                              logging._nameToLevel.keys())),
+                        type=click.Choice(
+                            list(map(str.lower, logging._nameToLevel.keys()))
```
znatty22:
```suggestion
                            [level_name.lower()
                             for level_name in logging._nameToLevel.keys()]
```

fiendish: lol sure

PR #283, opened 2019-04-30
🚸 Nicer handling of missing(stubbed) data

fiendish:
gives things like `DataIngestPipeline - ERROR - TransformStage received no input. Did the previous stage produce anything?`

### DIFF
--- a/kf_lib_data_ingest/etl/transform/transform.py
+++ b/kf_lib_data_ingest/etl/transform/transform.py
@@ -127,6 +127,12 @@ class TransformStage(IngestStage):
         except TypeError as e:
             raise InvalidIngestStageParameters from e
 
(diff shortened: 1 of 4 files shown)


kf_lib_data_ingest/etl/transform/transform.py
```diff
@@ -264,8 +264,13 @@ def _run(self, data_dict):
         :returns target_instances: dict - keyed by target concept -
         of lists containing dicts - representing target concept instances.
         """
-
-        target_instances = self._do_transform(data_dict)
+        if data_dict:
```
znatty22: This seems redundant with _validate_run_parameters. Could move this type of check/messaging there

PR #285, opened 2019-04-30
♻️ Update cat study

znatty22:
Cat study is used in integration test with study creator. Made it more realistic


fiendish:
This works with the current state of #254

PR #292, opened 2019-05-03
🚚 Mv network funcs into network utils

znatty22:


fiendish:
probably get_open_api_v2_schema could be separate from the session maker?

PR #293, opened 2019-05-03
🔊 Log release version and commit sha

znatty22:
Log looks like:
```
2019-05-03 16:24:51,174 - DataIngestPipeline - INFO - -- Ingest Params --
{'auto_transform': False,
 'dry_run': True,
 'ingest_package_dir': '/Users/singhn4/Projects/kids_first/kf-lib-data-ingest/my_study',
 'log_dir': '/Users/singhn4/Projects/kids_first/kf-lib-data-ingest/my_study/logs',
 'log_level_name': 'debug',
 'overwrite_log': None,
 'target_api_config_path': '/Users/singhn4/Projects/kids_first/kf-lib-data-ingest/kf_lib_data_ingest/target_apis/kids_first.py',
 'target_url': 'http://localhost:5000',
 'use_async': False,
 'version': '0.3.0'}
```


kf_lib_data_ingest/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = '0.3.0'
```
fiendish: Can this be populated directly from git during build/install instead of hard-coded?

Dockerfile
```diff
@@ -4,16 +4,14 @@ ENV KF_INGEST_APP_MODE production
...
-
-RUN pip install --upgrade pip && \
-    pip install --no-cache-dir -e .
-
 RUN apt-get update && apt-get install -y jq wget && \
     wget -q -O vault.zip https://releases.hashicorp.com/vault/1.0.3/vault_1.0.3_linux_amd64.zip && \
     unzip vault.zip && \
     mv vault /usr/local/bin
 
 COPY . /app
```
fiendish: please put a trailing slash on the destination directory of a copy

kf_lib_data_ingest/__init__.py
znatty22: What do you mean?

fiendish: I mean that this file should be autogenerated.

znatty22: Sure, but probably in another PR. Autogenerating this file should probably be part of the release making process, bc this file should only be updated when we want to make a new release

tests/test_network_utils.py
```diff
@@ -0,0 +1,37 @@
...
+    caplog.set_level(logging.DEBUG)
+
+    # Test successful get
+    assert get_tag_commit_sha(VERSION)
+    assert 'Got tags' in caplog.text
+
+    # Test get tag that doesn't exist
+    with pytest.raises(Exception):
+        try:
+            get_tag_commit_sha(VERSION + 1)
```
fiendish: Isn't VERSION a string?

fiendish: Isn't VERSION a string?

tests/test_network_utils.py
```diff
@@ -0,0 +1,37 @@
...
+    Test get_tag_commit_sha
+    """
+    caplog.set_level(logging.DEBUG)
+
+    # Test successful get
+    assert get_tag_commit_sha(VERSION)
+    assert 'Got tags' in caplog.text
+
+    # Test get tag that doesn't exist
+    with pytest.raises(Exception):
```
fiendish: This is masking your str+int error.

tests/test_network_utils.py
```diff
@@ -0,0 +1,37 @@
...
+        try:
+            get_tag_commit_sha(VERSION + 1)
+        except Exception as e:
+            assert 'Could not find tag' in caplog.text
```
fiendish: This needs to be outside of the with.

tests/test_network_utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Test get tag when status_code != 200
+    with requests_mock.Mocker(real_http=True) as m:
+        endpoint = f'{PROJECT_GITHUB_URL}/tags'
+        m.get(endpoint, status_code=400)
+        with pytest.raises(Exception):
+            try:
+                get_tag_commit_sha('foo')
+            except Exception as e:
+                assert 'Could not fetch tags' in caplog.text
```
fiendish: This needs to be outside of the with.

tests/test_network_utils.py
znatty22: woops yes

tests/test_network_utils.py
```diff
@@ -0,0 +1,31 @@
...
+    caplog.set_level(logging.DEBUG)
+
+    # Test successful get
+    assert get_tag_commit_sha(VERSION)
+    assert 'Got tags' in caplog.text
+
+    # Test get tag that doesn't exist
+    with pytest.raises(Exception) as e:
+        get_tag_commit_sha('foo')
+        assert 'Could not find tag' in str(e)
```
fiendish: unindent the assert

tests/test_network_utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # Test get tag when status_code != 200
+    with requests_mock.Mocker(real_http=True) as m:
+        endpoint = f'{PROJECT_GITHUB_URL}/tags'
+        m.get(endpoint, status_code=400)
+        with pytest.raises(Exception) as e:
+            get_tag_commit_sha('foo')
+            assert 'Could not fetch tags' in str(e)
```
fiendish: unindent the assert

tests/test_network_utils.py
znatty22: But its checking for the message in the exception msg

tests/test_network_utils.py
znatty22: Same as ^

tests/test_network_utils.py
fiendish: Doesn't matter. If the assert fails it will get swallowed.

fiendish:
```suggestion
        emsg = str(e)
assert 'Could not find tag' in emsg
```

fiendish: I guess you want something like:
```
with pytest.raises(Exception):
      try:
          get_tag_commit_sha('foo')
      except Exception as e:
          emsg = str(e)
          raise
assert 'Could not find tag' in emsg
```

znatty22:
> This won't work for the reasons explained above. The assert needs to be outside the with.

Welp, then we have a lot of other places in the test code where we need to fix this...

kf_lib_data_ingest/tag_commit_cache.json
```diff
@@ -0,0 +1,5 @@
+{
+    "0.1.0": "502587d550fdc9ba1a7adc68a836a02b78c0a0f7",
+    "0.2.0": "271f03ff7b0b9477017dd837d9ae4c272cac89e5",
+    "0.3.0": "f1a50035c3d463c78f13abc3b0cb353afe200c0b"
+}
```
fiendish: Does this mean you want to update this file every release?

kf_lib_data_ingest/network/utils.py
```diff
@@ -245,3 +250,58 @@ def get_open_api_v2_schema(url, entity_names,
...
+        tag_commit_cache = read_json(tag_commit_cache_file)
+        commit_sha = tag_commit_cache.get(tag_name)
+        if commit_sha:
+            return commit_sha
+        else:
+            logger.debug(f'Could not find tag {tag_name} in tag commit cache')
+    else:
+        logger.debug('Tag commit cache does not exist yet')
+
+    # Try to cet commit from github API
```
fiendish: For completeness, I think we should try local git commands first before hitting the github api.

kf_lib_data_ingest/tag_commit_cache.json
fiendish: This is now redundant with __version__ = '0.3.0' in __init__.py, since we can look at the largest tag.

kf_lib_data_ingest/network/utils.py
```diff
@@ -245,3 +250,58 @@ def get_open_api_v2_schema(url, entity_names,
             print(err)
 
     return output
+
+
+def get_tag_commit_sha(tag_name,
```
fiendish: If you can write the tag in a file, why can't you just write down the commit hash for the tag and not have to go fetch it?

znatty22: True, we could do away with this function if something else (maybe the release process) is able to generate an updated release tag, commit hash file when a new release is made. Then this file does become redundant with `__init__.py` bc we can look at the latest/last entry in the file.

kf_lib_data_ingest/tag_commit_cache.json
znatty22: Bc someone may pip install version 0.2.0 or a version that is not the latest version. You still need to get the commit hash for that version somehow

kf_lib_data_ingest/network/utils.py
znatty22: Its not a standard, but I read that people typically store the version of a python package in __version__. I've seen that stored in the package's __init__.py, or some other file. Maybe for us, since we will always want to know the commit hash as well as the version, we should just store the mapping file of the tags and hashes?

znatty22: > Does this mean you want to update this file every release?

Yea, probably (and then as I mentioned above, we would get rid of `__init__.py` w/ the version

kf_lib_data_ingest/tag_commit_cache.json
fiendish: I propose making the version file (\_\_init\_\_.py) always just have both the tag and also the commit hash for the current version and nothing else, get rid of this file, and get rid of the lookup function. If someone installs an older version, their version file would include the appropriate hash for that version. We can work on replacing manual updates of the version file with automated ones in another pr for the release process.

fiendish:
I think this does too much. I propose that you:
 
store the current tag and also the commit hash for that tag in \_\_init\_\_.py.
remove the lookup function
remove the cache file

You don't need to store previous tags, because any install of a previous version will have the hash for that tag  in the file.

znatty22:
@fiendish  I just realized something. We will only be able to update the commit hash in the package's `__init__.py` after the release PR is merged bc the commit hash points to the tagged release commit. Which means we won't be able to autogenerate an `__init__.py` with both the release tag and the commit sha during the release making process. If we keep what we have, then we'll always need another PR to update `__init__.py` with the commit sha for the release tag in `__version__`.

I think this is why I opted to lookup the commit sha for the tag and cache it after I've looked it up the first time. I just forgot why I did it 😆 Personally, I'd rather not have to make an additional PR to store the commit hash. I'd rather have something to get it for us automatically. We could still cache it inside `__init__.py` in `__commit_hash__` attribute too so we don't need that extra JSON file I had before.

What do you think?

kf_lib_data_ingest/etl/ingest_pipeline.py
```diff
@@ -114,6 +118,12 @@ def __init__(
         kwargs = {arg: values[arg] for arg in args[2:]}
         # Don't log anything that might have secrets!
         kwargs.pop('auth_configs', None)
+
+        # Log ingest package dir, ingest lib version, and commit hash
+        kwargs['ingest_package_dir'] = self.ingest_config_dir
+        kwargs['version'] = VERSION
+        kwargs['commit'] = COMMIT_SHA
```
fiendish: Why are these thrown into kwargs instead of using a new call to logger?

fiendish:
> @fiendish I just realized something. We will only be able to update the commit hash in the package's `__init__.py` after the release PR is merged bc the commit hash points to the tagged release commit. Which means we won't be able to autogenerate an `__init__.py` with both the release tag and the commit sha during the release making process. If we keep what we have, then we'll always need another PR to update `__init__.py` with the commit sha for the release tag in `__version__`.

There's no reason why the hash needs to be at the tag as opposed to at the last commit before the tag.

znatty22:
@fiendish Ok, then I propose we just don't record the commit hash

fiendish:
@znatty22 well, we _could_ try a local git call for the current commit hash and log that if it doesn't error

kf_lib_data_ingest/etl/ingest_pipeline.py
znatty22: Idk, I just did it. Could be its own logged message. Is that preferred?

PR #298, opened 2019-05-08
👌 PR suggestions

znatty22:


fiendish:
Why append something to then not use it?

znatty22:
Idk, I just thought the code for this approach was just more intuitive. I look at the code and go to the process leaf node block and see very clearly how include_root is used - first element of the property path is being stripped off. In the original PR its more difficult to understand how that skip_first flag is being used.

fiendish:
yeah ok sure

PR #303, opened 2019-05-15
🐛 Fix numeric_to_str - vals that equate to 0 should return '0'

### LINKED ISSUE #302: numeric_to_str should turn 0.0 into '0' not '0.0'
I expected that [numeric_to_str](https://github.com/kids-first/kf-lib-data-ingest/blob/c206287e0fc4f65d411cbeb97c61b9aaeaad7f0b/kf_lib_data_ingest/common/misc.py#L90) would convert the float `0.0` into the string `'0'` not `'0.0'`, because it turns `1.0` into `'1'`.

Is this the desired behavior? If so, why?

znatty22:
Fixes #302

### DIFF
--- a/kf_lib_data_ingest/common/misc.py
+++ b/kf_lib_data_ingest/common/misc.py
@@ -124,19 +124,29 @@ def numeric_to_str(val, replace_na=False, na=None):
 
     val = str(val).strip()
     if val != '':
-        # Don't automatically change anything with leading zeros, scientific
-        # notation, or underscores (I don't care what PEP 515 says).
-        if (val[0] == '0') or (not re.fullmatch(r'[\d.]+', val)):
-            return val
+        # Try downcasting val
+        i_val = None
+        f_val = None
         try:
             f_val = float(val)
             i_val = int(f_val)
-            if i_val == f_val:
-                return str(i_val)
-        except Exception:
+        except ValueError:
             pass
+
+        # Don't automatically change anything with leading zeros
+        # (except something that equates to int 0), scientific
+        # notation, or underscores (I don't care what PEP 515 says).
+        if (i_val != 0) and ((val[0] == '0') or
+                             (not re.fullmatch(r'[\d.]+', val))):
+            return val
+
+        # Return str version of downcasted val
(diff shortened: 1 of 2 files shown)


kf_lib_data_ingest/common/misc.py
```diff
@@ -124,19 +124,32 @@ def numeric_to_str(val, replace_na=False, na=None):
...
             pass
+
+        # Don't automatically change anything with leading zeros
+        # (except something that equates to int 0), scientific
+        # notation, or underscores (I don't care what PEP 515 says).
+        if (
+            not (i_val == 0) and  # does not equate to int 0
+                (val[0] == '0') or  # starts with leading zeros
+                (not re.fullmatch(r'[\d.]+', val))  # scientific notation
+        ):
```
fiendish: The operator precedence is confusing on this conditional. Can you put parentheses in to clarify desired order of operations?

znatty22: Haha oops my bad, thats terrible

kf_lib_data_ingest/common/misc.py
```diff
@@ -124,19 +124,29 @@ def numeric_to_str(val, replace_na=False, na=None):
...
+
+        # Don't automatically change anything with leading zeros
+        # (except something that equates to int 0), scientific
+        # notation, or underscores (I don't care what PEP 515 says).
+        if (i_val != 0) and ((val[0] == '0') or
+                             (not re.fullmatch(r'[\d.]+', val))):
+            return val
+
+        # Return str version of downcasted val
+        if i_val == f_val:
```
fiendish: I think moving this means doing extra work needlessly in the numeric case.

tests/test_misc.py
```diff
@@ -20,3 +21,5 @@ def test_numeric_to_str():
 
     assert df1['b'][0] == '1_1'
     assert df1['b'][1] == 'ab'
+    assert df1['a'][0] == '1'
+    assert df1['a'][3] == '0'
```
fiendish: please add some float checks too

PR #308, opened 2019-05-20
♻️ Don't defer reading output

fiendish:
Punting the read_output like that means you're not actually reading output from the disk and will always pass the wrong stage's output if you skip anything except the first stage.


znatty22:
> Punting the read_output like that means you're not actually reading output from the disk and will always pass the wrong stage's output if you skip anything except the first stage.

Not sure what you mean by that. I'm iterating through stages in order because we're using `_iterate_stages` in the for loop. If a stage is not in the set the user supplied, then we skip execution all together but save a pointer to the skipped stage. Then whenever we do execute a stage, if the real-time output doesn't exist, we just read it in from the previous stage which we saved a pointer to.

fiendish:
> I'm iterating through stages in order because we're using `_iterate_stages` in the for loop.

Follow the `output` variable when you skip a stage. Where does the data come from? Where does it go?

znatty22:
ok let's break down your example:
```
kidsfirst ingest my_study --log_level=debug --stages=l

stages_to_run = {LoadStage}

index = 0 
for stage in _iterate_stages (ExtractStage, TransformStage, LoadStage):
    stage = ExtractStage
    1. Is stage in stages_to_run?
        No:
            previous_stage = ExtractStage
            continue for loop

index = 1
for stage in _iterate_stages (ExtractStage, TransformStage, LoadStage):
    stage = TransformStage
    1. Is stage in stages_to_run?
        No:
            previous_stage = TransformStage
            continue for loop

index = 2
for stage in _iterate_stages (ExtractStage, TransformStage, LoadStage):
    stage = LoadStage
    1. Is stage in stages_to_run?
        Yes:
            previous_stage still equal to TransformStage

    2. Does output exist?
        No:
            Use previous_stage.read_output()

    3. Execute LoadStage with ^ output
```

And here is a log of some print statements I put in to make sure this is happening:
```
--------------- Iteration 0 ---------------

Skipping execution of <class 'kf_lib_data_ingest.etl.extract.extract.ExtractStage'>
Save Previous Stage: <class 'kf_lib_data_ingest.etl.extract.extract.ExtractStage'>

--------------- Iteration 1 ---------------

Skipping execution of <class 'kf_lib_data_ingest.etl.transform.transform.TransformStage'>
Save Previous Stage: <class 'kf_lib_data_ingest.etl.transform.transform.TransformStage'>

--------------- Iteration 2 ---------------

Loading previously cached output from GuidedTransformStage
Executing stage <class 'kf_lib_data_ingest.etl.load.load.LoadStage'>
```

fiendish:
```ok let's break down your example:

kidsfirst ingest my_study --log_level=debug --stages=l
```

Try `kidsfirst ingest my_study --log_level=debug --stages=el`

fiendish:
> FileNotFoundError: Error reading LoadStage output. 

Also, the rule was that leaving out a stage means that output from a previous run must exist. If you want the rule to only apply to stages that don't have a to-run stage immediately following them, more refactor will be needed.

Issue #313, opened 2019-05-23
Method clean_up_df should only convert numeric types to str


znatty22:
The `clean_up_df` method currently tries to convert every value in a DataFrame to a string. It should only do this for numeric types, not lists or dicts. This comes into play when trying to load data:

```
2019-05-23 10:10:39,990 - DataIngestPipeline - ERROR - {"_status":{"code":400,"message":"could not create genomic_file: {'hashes': ['Not a valid mapping type.'], 'urls': ['Not a valid list.']}}

2019-05-23 10:10:39,989 - LoadStage - DEBUG - Request body:
{'availability': 'Immediate Download',
 'data_type': 'Aligned Reads',
 'external_id': 's3://kf-seq-data-bcm/chung/HKT73CCXX-5.hgv.bam',
 'file_format': 'bam',
 'file_name': 'HKT73CCXX-5.hgv.bam',
 'hashes': "{'md5': '98ebe560095b231a3050812cc58a8b84-8100'}",
 'is_harmonized': 'False',
 'reference_genome': 'hs37d5',
 'sequencing_experiment_id': None,
 'size': '67939600949',
 'urls': "['s3://kf-seq-data-bcm/chung/HKT73CCXX-5.hgv.bam']"}
```
The field `hashes` should be a dict and field `urls` should be a list since that's what these fields were mapped to in the extract configs. However, because `clean_up_df` applies `numeric_to_str` to every value in the data, these types are lost.

A check should be added to `clean_up_df` to apply `numeric_to_str` to values are are not of type `list` or `dict`.

Additionally, changes must be made to stage's `_post_concept_discovery` since now the DataFrame data may have strings, lists, and dicts and the current implementation will error.

fiendish:
'98ebe560095b231a3050812cc58a8b84-8100' is not a valid md5

fiendish:
🤔

What if the loader just runs ast.literal_eval on the values to check for dicts or a lists?

fiendish:
This specific issue also goes away when https://github.com/kids-first/kf-api-dataservice/issues/487 is fixed

znatty22:
> This specific issue also goes away when kids-first/kf-api-dataservice#487 is fixed

This is true, but converting every value to a string still isn't the right thing to do. 

Now I'm even wondering whether we should be converting numeric types to strings in `clean_up_df`. I don't like that my source data has one set of types but then the data that my extract config operations are operating on has a different set of types. 

[...]

PR #314, opened 2019-05-23
🐛🔧 Fix/update genomic file std and target concept configuration

znatty22:


kf_lib_data_ingest/common/constants.py
```diff
@@ -15,6 +15,15 @@ class COMMON:
     MULTIPLE = "Multiple"
 
 
+class FILE:
+    class HASH:
+        MD5 = 'md5'
```
fiendish:
```suggestion
        MD5 = 'MD5'
```

kf_lib_data_ingest/common/constants.py
```diff
@@ -15,6 +15,15 @@ class COMMON:
...
+        SHA = 'sha'
+        SHA256 = 'sha256'
```
fiendish:
```suggestion
        SHA256 = 'SHA-256'
```

kf_lib_data_ingest/common/constants.py
```diff
@@ -15,6 +15,15 @@ class COMMON:
...
+        SHA512 = 'sha512'
```
fiendish:
```suggestion
        SHA512 = 'SHA-512'
```

kf_lib_data_ingest/common/constants.py
```diff
@@ -15,6 +15,15 @@ class COMMON:
...
+        CRC = 'crc'
```
fiendish: CRC isn't one thing, and none of them are useful anyway

```suggestion
        S3_ETAG = S3 ETag'
```

kf_lib_data_ingest/target_apis/kids_first.py
```diff
@@ -154,23 +154,19 @@
...
             "is_harmonized": CONCEPT.GENOMIC_FILE.HARMONIZED,
-            "paired_end": CONCEPT.READ_GROUP.PAIRED_END,
-            "hashes": CONCEPT.GENOMIC_FILE.UNIQUE_KEY,
-            "size": CONCEPT.GENOMIC_FILE.UNIQUE_KEY,
-            "urls": CONCEPT.GENOMIC_FILE.UNIQUE_KEY,
-            "acl": CONCEPT.GENOMIC_FILE.UNIQUE_KEY,
-            "reference_genome": CONCEPT.GENOMIC_FILE.UNIQUE_KEY,
+            "hashes": CONCEPT.GENOMIC_FILE.HASH,
+            "size": CONCEPT.GENOMIC_FILE.SIZE,
+            "urls": CONCEPT.GENOMIC_FILE.URL,
```
fiendish: doesn't urls need to be a list?

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -7,9 +7,9 @@
 class FileMixin(object):
     ETAG = None
```
fiendish:
```suggestion
```

kf_lib_data_ingest/target_apis/kids_first.py
znatty22: Yes it does but we don't have a way to define types in the target schema yet. That would be kind of a big change and require some design + discussion. 

For right now, I propose we make a fix for https://github.com/kids-first/kf-lib-data-ingest/issues/313 and then the ingest pkg dev can map this column to a python list containing their URL like in [this example](https://github.com/kids-first/kf-ingest-packages/blob/26f8c1981278eb501196f50a840cbe03eb4155ac/kf_ingest_packages/common/extract_configs/s3_object_info.py#L127) and then the value will go directly into `urls` during the standard to target transformation.

fiendish: then change the concept name to URLS plural

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -7,9 +7,9 @@
...
     SIZE = None
-    DATA_TYPE = None
+    HASH = None
```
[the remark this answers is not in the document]
znatty22: Yea I think this will have to be a dict bc a file can have more than one hash and we have to know which value goes with which type. Right now all we can do is call it `HASHES` and then leave it to the developer of the extract config to map a dict to it like in [this example](https://github.com/kids-first/kf-ingest-packages/blob/26f8c1981278eb501196f50a840cbe03eb4155ac/kf_ingest_packages/common/extract_configs/s3_object_info.py#L145) 

I think we definitely need to sit down and discuss design of both concept schema and target API config schema bc neither one of them account for types

kf_lib_data_ingest/common/constants.py
znatty22: Haha I stole crc from indexd bc its one of their options for hashes

kf_lib_data_ingest/target_apis/kids_first.py
znatty22: Ok, then in that case, what do you think about: `HASHES_DICT` and `URL_LIST`? Bc if we name them `HASHES` and `URLS` how will someone know one is meant to be a dict vs a list

fiendish: ok

fiendish: this is all hopefully temporary anyway until gen3 usage is fixed

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -5,11 +5,11 @@
 
 
 class FileMixin(object):
-    ETAG = None
     SIZE = None
-    DATA_TYPE = None
+    URL = None
```
znatty22: I included URL and HASH, HASH_TYPE in the FileMixin and then URL_LIST and HASH_DICT in the GENOMIC_FILE concept because we shouldn't only cater to the types/structure of the Dataservice. This way we can use URL for a file path or file url and we can use URL_LIST for a list of filepaths or urls. Sort of similar for hashes

fiendish: I don't think that makes any sense.

First, you don't have any way to convert URL into URL_LIST or HASH into HASH_DICT, which means that all of your URL and HASH values are currently being thrown away.

> I thought we shouldn't only cater to the types/structure of the Dataservice

It has nothing to do with the dataservice. A file may always be stored in multiple locations and definitely always has infinitely many hash values, no matter where you choose to store the information about your files.

znatty22: Ok, so then if we only have URL_LIST and HASH_DICT, then what should go in here:
https://github.com/kids-first/kf-lib-data-ingest/blob/daf0a280dd43c46a8da6368355cf3e5de65d2883/tests/data/test_study/extract_configs/simple_tsv_example2.py#L39

Should it the mapping function still return a string even though will be mapped to something called `CONCEPT.GENOMIC_FILE.URL_LIST`?

fiendish: return a list

tests/data/test_study/extract_outputs/simple_tsv_example2_output.tsv
```diff
@@ -1,4 +1,4 @@
-index	BIOSPECIMEN|ALIQUOT_ID	GENOMIC_FILE|FILE_PATH	GENOMIC_FILE|FILE_NAME	BIOSPECIMEN|ID
+index	BIOSPECIMEN|ALIQUOT_ID	GENOMIC_FILE|URL_LIST	GENOMIC_FILE|FILE_NAME	BIOSPECIMEN|ID
 0	4	FLDR0/2.bam	2.bam	4
```
fiendish: these now need to be lists of strings

tests/data/test_study/extract_outputs/unsimple_xlsx_example1_output.tsv
```diff
@@ -1,4 +1,4 @@
-index	BIOSPECIMEN|ALIQUOT_ID	BIOSPECIMEN|ID	GENOMIC_FILE|FILE_PATH	BIOSPECIMEN|ANALYTE
+index	BIOSPECIMEN|ALIQUOT_ID	BIOSPECIMEN|ID	GENOMIC_FILE|URL_LIST	BIOSPECIMEN|ANALYTE
 0	4	4	s3://foo/bar/vgrnqzhvrudvz3z/mn0ql.bam	DNA
```
fiendish: these now need to be lists of strings

tests/data/test_study/extract_outputs/simple_tsv_example2_output.tsv
znatty22: oh oops lol forgot to add that part

tests/data/test_study/extract_configs/simple_tsv_example2.py
```diff
@@ -34,10 +34,10 @@
     ),
     row_map(
         m=lambda row: row['storage_dir'] + '/' + row['bam_file_name'],
```
fiendish: these need to be lists now
```suggestion
        m=lambda row: [row['storage_dir'] + '/' + row['bam_file_name']],
```

PR #315, opened 2019-05-25
✨ Add grouping of Splits

### LINKED ISSUE #301: Support non-cartesian product row splitting
Our current row splitting operation basically produces a set of rows that reflect the cartesian product of col A and col B. 

This is fine for some situations but I also need a way to turn this: 

| col A | col B |
-------| -----|
|congenital diaphragmatic hernia; Esophageal stricture;  foo| HP:0000776; HP:0002043|

into: 

| col A | col B |
-------| -----|
|congenital diaphragmatic hernia | HP:0000776|
|Esophageal stricture | HP:0002043|
|foo | None |

fiendish:
This lets you specify a group ID for Split objects, such that two
Splits on the same row with the same group ID will be applied
simultaneously in parallel rather than sequentially.

closes #301

### DIFF
--- a/kf_lib_data_ingest/common/pandas_utils.py
+++ b/kf_lib_data_ingest/common/pandas_utils.py
@@ -21,64 +21,106 @@ class Split:
 
     Replaces lists with a Split object containing the list.
     """
-
-    def __init__(self, things):
+    def __init__(self, things, group=None):
         assert_safe_type(things, list)
         self.things = things
+        self.group = group
 
 
 def split_df_rows_on_splits(df):
     """
     Take a DataFrame and split any cell values that contain Splits into
-    multiple rows.
+    multiple rows. Generally, multiple Splits on the same row will multiply to
+    produce the cartesian product of the values in those Splits. However, if
+    Splits are assigned a `group` value, then Splits on the same row in the
+    same group will be linked to each other so that they do not form a
+    cartesian product.
 
     e.g.:
 
     a row that looks like...
 
-    {'a': Split([1, 2]), 'b': Split([[3, 4])}
+    {'a': Split([1, 2], 1), 'b': Split([3, 4, 5], 1), 'c': Split([6, 7])}
 
     ...will become...
 
     [
-        {'a': 1, 'b': 3},
-        {'a': 1, 'b': 4},
-        {'a': 2, 'b': 3},
-        {'a': 2, 'b': 4}
+        {'a': 1,    'b': 3, 'c': 6},
+        {'a': 2,    'b': 4, 'c': 6},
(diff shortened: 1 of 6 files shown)


fiendish:
> How will this work with `split_df_rows_on_delims` now?

It won't. And nothing in the repository uses that function anyway.

znatty22:
> It won't. And nothing in the repository uses that function anyway.

It could though. You could change the method to this: 
```
def split_df_rows_on_delims(df, cols: list = None, delimiters: list = None, cartesian=True):
    """
    Split row into multiple rows based on delimited strings in df[col]
    for all columns in cols
    """
    group = int(cartesian) or None
    cols = cols or df.columns
    for i, row in df.iterrows():
        for col in cols:
            val = multisplit(row[col], delimiters)
            if len(val) > 1:
                row[col] = Split(val, group=group)
            else:
                row[col] = val[0]
    return split_df_rows_on_splits(df)
```

fiendish:
lol what is `int(cartesian)` doing if you can only set one group

kf_lib_data_ingest/common/pandas_utils.py
```diff
@@ -21,64 +21,106 @@ class Split:
...
+    df, cols: list = None, delimiters: list = None, cartesian: bool = True
+):
     """
     Split row into multiple rows based on delimited strings in df[col]
     for all columns in cols
     """
+    assert_safe_type(cols, list)
+    assert_safe_type(delimiters, list)
+    assert_safe_type(cartesian, bool)
+    group = None if cartesian else 'group_1'
```
znatty22: How come its `'group_1'` and not `1`?

kf_lib_data_ingest/common/pandas_utils.py
```diff
@@ -21,50 +21,84 @@ class Split:
...
     e.g.:
 
     a row that looks like...
 
-    {'a': Split([1, 2]), 'b': Split([[3, 4])}
+    {'a': Split([1, 2], 1), 'b': Split([3, 4, 5], 1), 'c': Split([5, 6])}
 
     ...will become...
 
     [
```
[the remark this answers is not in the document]
znatty22: ?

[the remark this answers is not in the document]
fiendish: oops

kf_lib_data_ingest/common/pandas_utils.py
fiendish: because it doesn't matter what it is

fiendish: made it 1 now

PR #316, opened 2019-05-25
🐛 Load objects instead of object-like strings

### LINKED ISSUE #313: Method clean_up_df should only convert numeric types to str

fiendish:
Will recurse into lists and dicts to clean their contents for comparison across stages and then revert to their original native type.

closes #313

### DIFF
--- a/kf_lib_data_ingest/common/misc.py
+++ b/kf_lib_data_ingest/common/misc.py
@@ -1,3 +1,4 @@
+import ast
 import datetime
 import importlib
 import inspect
@@ -87,7 +88,7 @@ def iterate_pairwise(iterable):
     return zip(a, b)
 
 
-def numeric_to_str(val, replace_na=False, na=None):
+def convert_to_downcasted_str(val, replace_na=False, na=None):
     """
     Converts values to stripped strings while collapsing downcastable floats.
 
@@ -116,6 +117,16 @@ def numeric_to_str(val, replace_na=False, na=None):
 
     :return: new representation of `val`
     """
+    if isinstance(val, list):
+        # make hashable without changing style or losing comparability
+        return str(sorted(
+            convert_to_downcasted_str(v) for v in val
+        ))
+    if isinstance(val, dict):
+        # make hashable without changing style or losing comparability
+        return str(dict(sorted(
+            (k, convert_to_downcasted_str(v)) for k, v in val.items()
+        )))
     if isnull(val):
         if replace_na:
             return na
@@ -150,6 +161,34 @@ def numeric_to_str(val, replace_na=False, na=None):
     return val
 
 
+def str_to_obj(var):
+    """
+    Convert a string that looks like a list, dict, or tuple back into its
+    native object form.
+    """
+    if not isinstance(var, str):
+        return var
+    if var.startswith(("[", "{", "(")):
+        try:
+            return ast.literal_eval(var)
+        except Exception:
+            pass
+    return var
+
+
+def recover_containers_from_df_strings(df):
+    """
+    Undo one bit of necessary madness imposed by clean_up_df where lists and
+    dicts (both unhashable) are sorted and then converted to strings for
+    safekeeping. This finds strings that look like lists, dicts, or tuples and
+    converts them back to their native forms.
+
+    :param df: a pandas DataFrame
+    :return: Dataframe with object-like strings converted to native objects
+    """
+    return df.applymap(str_to_obj)
+
+
 def clean_up_df(df):
     """
     We can't universally control which null type will get used by a data
@@ -169,7 +208,7 @@ def clean_up_df(df):
     """
 
     return df.applymap(
-        lambda x: numeric_to_str(x, replace_na=True, na=None)
+        lambda x: convert_to_downcasted_str(x, replace_na=True, na=None)
     )
 
 
--- a/kf_lib_data_ingest/etl/load/load.py
+++ b/kf_lib_data_ingest/etl/load/load.py
@@ -13,7 +13,7 @@ from requests import RequestException
 from sqlite3worker import Sqlite3Worker, sqlite3worker
 
 from kf_lib_data_ingest.common.errors import InvalidIngestStageParameters
-from kf_lib_data_ingest.common.misc import multisplit
+from kf_lib_data_ingest.common.misc import multisplit, str_to_obj
 from kf_lib_data_ingest.common.stage import IngestStage
 from kf_lib_data_ingest.network.utils import requests_retry_session
 from kf_lib_data_ingest.config import DEFAULT_ID_CACHE_FILENAME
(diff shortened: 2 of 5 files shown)


fiendish:
I decided that the deconversion of the object types should be part of load and not transform, because it's really a transport-related issue (json body) not a transforming issue.

kf_lib_data_ingest/common/misc.py
```diff
@@ -87,7 +88,7 @@ def iterate_pairwise(iterable):
     return zip(a, b)
 
 
-def numeric_to_str(val, replace_na=False, na=None):
+def convert_to_str(val, replace_na=False, na=None):
```
znatty22: Maybe this should be called `convert_to_downcasted_str` so that someone is aware that there is something more than just conversion to strs going on.

kf_lib_data_ingest/common/misc.py
```diff
@@ -116,6 +117,16 @@ def numeric_to_str(val, replace_na=False, na=None):
 
     :return: new representation of `val`
     """
+    if isinstance(val, list):
+        # make hashable without changing style or losing comparability
+        return str(sorted(
```
znatty22: Why do you need to sort it? What if the order of the list mattered in the source data?

kf_lib_data_ingest/common/misc.py
```diff
@@ -116,6 +117,16 @@ def numeric_to_str(val, replace_na=False, na=None):
...
+            convert_to_str(v) for v in val
+        ))
+    if isinstance(val, dict):
+        # make hashable without changing style or losing comparability
+        return str(dict(sorted(
```
znatty22: Same comment as above since Python 3.7 maintains order of keys in dicts

kf_lib_data_ingest/etl/load/load.py
```diff
@@ -309,8 +309,11 @@ def _load_entity(self, entity_type, endpoint, entity_id, body, links):
             or self._get_target_id(entity_type, entity_id)
         )
 
+        # convert list/dict-like strings to their native forms and
         # don't send null attributes
-        body = {k: v for k, v in body.items() if not pandas.isnull(v)}
+        body = {
+            k: str_to_obj(v) for k, v in body.items() if not pandas.isnull(v)
```
znatty22: Hmm, I'm wondering whether this should be here or in transform stage. Idk what the best solution is but here are some thoughts ...

What if we wanted to develop an additional loader or someone wanted to do something with the transform stage's JSON output. 

In the former case, now the new loader also has to remember to do this `str_to_obj` conversion. 

In the latter case, its weird for someone to see list and dict type values in string form in a JSON file. If someone wanted to do something with this transform output, now they also have to know to `str_to_obj` on all values.

kf_lib_data_ingest/common/misc.py
fiendish: √

kf_lib_data_ingest/common/misc.py
fiendish: I think the order of the list will never actually matter. In practice, all lists in data should probably be sets.

kf_lib_data_ingest/common/misc.py
fiendish: but when the dict is converted to a string, you can't then compare equal dicts that were created with their keys in different order

kf_lib_data_ingest/etl/load/load.py
fiendish: > In the former case, now the new loader also has to remember to do this str_to_obj conversion.

I'm fine with this.

> In the latter case, its weird for someone to see list and dict type values in string form in a JSON file. If someone wanted to do something with this transform output, now they also have to know to str_to_obj on all values.

That's true. Hmm. I'll see what I can do about this.

kf_lib_data_ingest/common/misc.py
znatty22: Ok I see

kf_lib_data_ingest/common/misc.py
znatty22: Ok I see

kf_lib_data_ingest/etl/load/load.py
fiendish: So here's the catch...

dicts, lists, and sets aren't hashable, and that makes Python get very angry. You can't use them as dict keys, you can't put them in sets, and pandas very much can't merge on them. So no matter what you do, if you don't convert them to strings, somewhere deep inside you're going to have to convert them to _something_ hashable (frozenset, tuple, whatever) otherwise nothing works. Now those types definitely aren't compatible with JSON, so you're going to have to use our funny jsonpickled write_json and read_json functions from common.misc.

Therefore, I think at best we could add another flag to read_json for searching for and converting object-like strings when reading the file.

znatty22: Sigh, yea let's just stick with 1

fiendish:
> The method `numeric_to_str` preserves native types but now post run count analysis breaks
> 
> ```
> Traceback (most recent call last):
>   File "/Users/singhn4/Projects/kids_first/kf-lib-data-ingest/kf_lib_data_ingest/etl/ingest_pipeline.py", line 198, in run
>     output = stage.run()  # First stage gets no input
>   File "/Users/singhn4/Projects/kids_first/kf-lib-data-ingest/kf_lib_data_ingest/common/stage.py", line 146, in wrapper
>     r = func(instance, *args, **kwargs)
>   File "/Users/singhn4/Projects/kids_first/kf-lib-data-ingest/kf_lib_data_ingest/common/stage.py", line 173, in run
>     self.concept_discovery_dict = self._postrun_concept_discovery(output)
>   File "/Users/singhn4/Projects/kids_first/kf-lib-data-ingest/kf_lib_data_ingest/etl/extract/extract.py", line 410, in _postrun_concept_discovery
>     sk[val].add(data_file)
> TypeError: unhashable type: 'dict'
> ```
> 
> Unit tests for post run count analysis would catch this

that's what jenkins is

fiendish:
I know you have a new passing test study in another PR. That one should do for catching concept discovery errors.

PR #319, opened 2019-05-28
♻️ Refactor links in target api cfg 

znatty22:
Closes #318

PR #324, opened 2019-05-31
✨ Add family relationship to concept schema and kids first api config

### LINKED ISSUE #321: Add family relationship to concept schema and Kids First target API config
There is no way to load family relationships right now. Family relationship is missing from the concept schema and also from the KF target API config

znatty22:
Please review #319 first, will change base after that merges. 

Closes #321


znatty22:
Yea I've been using this along with the changes in #319 and it works 👌 so far

fiendish:
It looks like this got rebased on master?

PR #327, opened 2019-06-05
✨ Support resuming load from a given target ID or prefix

### LINKED ISSUE #317: Making loading resumable
Add the ability to dry-run until a particular KFID and then switch to actual loading.

fiendish:
Network traffic for loading is slow, so support aborting and resuming
later without resubmitting data that has already been submitted.

The given ID will get submitted. This is to support ID prefixes ('DG',
'FM', etc) where the expectation is to begin loading at the first one
without gratuitous checks and filtering.

closes #317

### DIFF
--- a/kf_lib_data_ingest/app/cli.py
+++ b/kf_lib_data_ingest/app/cli.py
@@ -87,7 +87,11 @@ def cli():
               default=False,
               is_flag=True,
               help='A flag specifying whether to only pretend to send data to '
-              'the target service')
+              'the target service. Overrides the resume_from setting.')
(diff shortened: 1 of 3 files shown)


fiendish:
> I think the user should be notified if the resume_from target was never found

I can add that, but the user will already know because it dry runs all the way to the end.

znatty22:
> I can add that message, but the user will already know because it dry runs all the way to the end instead of picking up at their target

True but I like something explicit that my eye is likely to catch if present (the little yellow warning emoji)

PR #329, opened 2019-06-06
🐛 Fix invalid use of pytest context managers

### LINKED ISSUE #300: oops, bad tests
per comments on https://github.com/kids-first/kf-lib-data-ingest/pull/293 we have tests that squelch assert failures inside `with` statements that should be outside those statements. Those need to be fixed.

fiendish:
When using pytest.raises as a context manager, normal context manager
rules apply, and the exception raised must be the final line in the
scope of the context manager. Lines of code after that within the scope
of the context manager will not be executed.

A few valid but verbose uses were also changed for consistency.

closes #300

### DIFF
--- a/tests/test_concept_graph.py
+++ b/tests/test_concept_graph.py
@@ -200,11 +200,11 @@ def test_to_from_dict():
     node_dict['somekey'] = 'foo'
     with pytest.raises(AttributeError) as e:
         ConceptNode.from_dict(node_dict)
-        assert 'concept_attribute_pair' in str(e)
+    assert 'somekey' in str(e.value)
 
     # Missing required key
     node_dict.pop('somekey')
     node_dict.pop('concept_attribute_pair')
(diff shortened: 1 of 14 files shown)


tests/test_concept_graph.py
```diff
@@ -200,11 +200,11 @@ def test_to_from_dict():
     node_dict['somekey'] = 'foo'
     with pytest.raises(AttributeError) as e:
         ConceptNode.from_dict(node_dict)
-        assert 'concept_attribute_pair' in str(e)
+    assert 'somekey' in str(e.value)
```
znatty22: Why not what it was before, `concept_attribute_pair`?

fiendish: because it's wrong, lol

znatty22: ohhh woops i didn't see line 200. I was like where did `somekey` come from lol

PR #330, opened 2019-06-06
♻️ Refactor BIOSPECIMEN concept -> BIOSPECIMEN, BIOSPECIMEN_GROUP

### LINKED ISSUE #310: Use ALIQUOT_ID as a Primary Key to CONCEPT.BIOSPECIMEN
For most of the X01 studies, a sample may be an aliquot. In CBTTC, however, an aliquot is an atomic level unit of biospecimen and a sample is a collection of aliquots. So, CBTTC uses the `external_aliquot_id` as a unique ID for a biospecimen entity. The current ingest extractor considers `BIOSPECIMEN.ID`, which is an `external_sample_id`, as a unique column and, in turn, accounting biospecimens against the ingest library will not be correct. For example, the Open DIPG (it is part of CBTTC) Chris Jones cohort that I am wrangling with the ingest library has 240 sample IDs and 410 aliquot IDs. If I use 410 as an expected count for `CONCEPT.BIOSPECIMEN`, it cannot pass the accounting.

znatty22:
Closes #310


kf_lib_data_ingest/target_apis/kids_first.py
```diff
@@ -182,8 +182,8 @@
             }
         ],
         'properties': {
-            'external_sample_id': CONCEPT.BIOSPECIMEN.UNIQUE_KEY,
-            'external_aliquot_id': CONCEPT.BIOSPECIMEN.ALIQUOT_ID,
+            'external_sample_id': CONCEPT.BIOSPECIMEN_GROUP.UNIQUE_KEY,
```
fiendish:
```suggestion
            'external_sample_id': CONCEPT.BIOSPECIMEN_GROUP.ID,
```

kf_lib_data_ingest/target_apis/kids_first.py
```diff
@@ -182,8 +182,8 @@
...
+            'external_aliquot_id': CONCEPT.BIOSPECIMEN.UNIQUE_KEY,
```
fiendish:
```suggestion
            'external_aliquot_id': CONCEPT.BIOSPECIMEN.ID,
```

tests/test_guided_transform.py
```diff
@@ -169,3 +171,43 @@ def f(df_dict):
...
+        'biospecimen': pd.DataFrame({
+            CONCEPT.BIOSPECIMEN.ID: ['b1', 'b2', 'b3'],
+            CONCEPT.BIOSPECIMEN_GROUP.ID: ['b1', 'b2', 'b3'],
+            CONCEPT.BIOSPECIMEN.ANALYTE: ['dna', 'rna', 'dna']
+        })
+    }
+    guided_transform_stage._validate_data(df_dict)
+    assert '✅ Valid biospecimen DataFrame' in caplog.text
+
+    # Test one column in missing
```
fiendish:
```suggestion
    # Test one column is missing
```

kf_lib_data_ingest/target_apis/kids_first.py
znatty22: Why not unique key?

kf_lib_data_ingest/target_apis/kids_first.py
znatty22: Why not unique key?

kf_lib_data_ingest/target_apis/kids_first.py
fiendish: Because that's not what unique key is for

fiendish: The dataservice _does_ need a field for what unique key is doing, but external ID isn't it

kf_lib_data_ingest/target_apis/kids_first.py
fiendish: ^

kf_lib_data_ingest/target_apis/kids_first.py
fiendish: unique keys should basically never go to the dataservice right now until it has a place to put our shenanigans that doesn't affect the portal

fiendish:
looking at the target service spec, I think every instance of UNIQUE_KEY that isn't a foreign key link should be replaced by ID

znatty22:
> looking at the target service spec, I think every instance of UNIQUE_KEY that isn't a foreign key link should be replaced by ID

But then what should go in `external_id` for diagnosis and phenotype observations? I realize we *should* put the ID assigned by the contributor into the external_id field and not mix in unique keys, but we are definitely doing that right now. We use the `external_id` for debugging/sanity checks. Again  I know we shouldn't be doing this but we are. I think until the dataservice has a separate place for unique keys we will continue doing that. If that's the case then can't we just keep storing unique keys in the external id field, does it hurt anything?

fiendish:
I edited my comment (I know I shouldn't do that) to include compound fields

fiendish:
> does it hurt anything?

Hm. That is the real question isn't it. Can you guarantee that it will always be the ID for things where the entity is identified by an ID?

fiendish:
I thought you said you were switching from BIOSPECIMEN_GROUP.ID to BIOSPECIMEN.GROUP_ID?

kf_lib_data_ingest/target_apis/kids_first.py
fiendish: This will affect the portal display. I don't know if that's acceptable.

znatty22: Lol ok so should I back out the changes I just made for BIOSPECIMENs UNIQUE KEY to be a compound of its sample group and aliquot?

fiendish: no. I think if allison says it shouldn't be on the portal then we change these to be the individual IDs

znatty22: I mean maybe we should just make them the individual ids since they are each capturing different ids?

fiendish: I guess we have this open question of exactly what the aliquot ID should contain. Should it be just the subcomponent or should it be a fully qualified ID more specific than the sample ID? TARGET data from GDC uses the latter.

fiendish: @allisonheath second question. should Aliquot ID always be a fully qualified identifier including the sample group component (like what GDC has with the TCGA barcodes)?

liberaliscomputing: Including extra components such as the same sample group into Aliquot ID seems somehow redundant (wouldn't hurt though). And other systems/applications such as PedcBioPortal may need to do additional data processing to parse out Aliquot ID from the compound ID.

PR #338, opened 2019-06-13
🐛 Remove errant layer from basic auth test config

### LINKED ISSUE #337: basic auth file retriever test is using an invalid config
This pytest complaint is because the test config is wrong.
```
test_file_retriever.py::test_get_web_w_auth[https://api.com/foo/1-basic-Selected `basic` authentication]
 kf-lib-data-ingest/venv/lib/python3.7/site-packages/requests/auth.py:44: DeprecationWarning: Non-string 
usernames will no longer be supported in Requests 3.0.0. Please convert the object you've passed in 
(None) to a string or bytes object in the near future to avoid problems.
    category=DeprecationWarning,
```
---
```Python
            'variable_names': {
                'username': os.environ.get('FOO_SERVICE_UNAME'),
                'password': os.environ.get('FOO_SERVICE_PW')
            }
```
should be
```Python
            'username': os.environ.get('FOO_SERVICE_UNAME'),
            'password': os.environ.get('FOO_SERVICE_PW')
```

fiendish:
The basic auth test wasn't making sure that the username and password actually got sent through. This fixes that.

closes #337

### DIFF
--- a/tests/test_file_retriever.py
+++ b/tests/test_file_retriever.py
@@ -1,10 +1,7 @@
-import os
+import base64
 import logging
-from urllib.parse import (
-    urljoin,
-    urlencode,
-    quote
-)
+import os
+from urllib.parse import quote, urlencode, urljoin
 
 import boto3
 import pytest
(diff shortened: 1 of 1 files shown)


tests/test_file_retriever.py
```diff
@@ -210,6 +205,10 @@ def test_get_web_w_auth(caplog, tmpdir, auth_configs, url, auth_type,
                 req_headers = m.request_history[0].headers
                 auth_header = req_headers.get('Authorization')
                 assert auth_header and 'Basic' in auth_header
+                username, password = base64.b64decode(
```
znatty22: Just wondering, does `requests_mock` always base64 encode the username, pw?

fiendish: It's not the mocker. RFC 2617 mandates base64 encoding. https://www.ietf.org/rfc/rfc2617.txt

znatty22: I know, but I was wondering whether `requests_mock` automatically does that for you or `requests` does it for you or if you as the developer must choose to follow RFC 2617 when sending the username, pw? Just wondering bc in our file retriever's web save method, we don't explicitly base64 encode username, pw and we prob should if its not done automatically.

znatty22: ok confirmed, requests does it for you

PR #343, opened 2019-06-18
✨ Error on inconsistent merge inputs

### LINKED ISSUE #340: merge_wo_duplicates should alert user when merge has to deal with inconsistent data
If you merge the following tables together on `ID` with regular pandas merge you get:

Table 1:

FILE_NAME | ID 
--- | ---
file1.bam | b1    
file2.bam | b2  
  
Table 2:

FILE_NAME | ID 
--- | ---
file_100.bam | b1    
file_200.bam | b2  

Result:

FILE_NAME_x | FILE_NAME_y | ID 
--- | --- | --
file1.bam| file_100.bam | b1    
file2.bam| file_200.bam | b2

If you merge these two tables together with `merge_wo_duplicates` you get something like:

FILE_NAME | ID 
--- | ---
file1.bam | b1    
file2.bam | b2  

because `merge_wo_duplicates` tries to eliminate the duplicate columns by collapsing them into one. 

I think in cases like this (both columns contain different and non-null values), we should raise an exception or at the very least log a warning to alert user that there is a major data inconsistency. 

This actually just happened to me because I incorrectly assumed that Chung samples sequenced by Broad were different than those sequenced by BCM. But it turned out that some samples were sequenced by both and I needed to include both.

fiendish:
closes #340


fiendish:
Ugh.

fiendish:
Ugh.

kf_lib_data_ingest/common/pandas_utils.py
```diff
@@ -327,6 +327,22 @@ def resolve_duplicates(df, suffixes):
                 if coll.endswith(l_suffix):
                     firstpart = coll.split(l_suffix)[0]
                     colr = firstpart + r_suffix
+                    inconsistent = (
```
znatty22: This logic doesn't quite work. If you try this:
```
df1 = pandas.DataFrame({
    'A': [1, 2, 2],
    'B': [3, 4, 5]
})
df2 = pandas.DataFrame({
    'A': [1, 2, 2],
    'B': [3, None, 6]
})
m = pandas.merge(df1, df2, on='A', how='outer')
```
it passes without exception even though a regular pandas merge shows inconsistent data:

A | B_x | B_y
-- | -- | -- 
1 | 3 | 3.0
2 | 4 | NaN
2 | 4 | 6.0
2 | 5 | NaN
2 | 5 | 6.0

I think you need something like this:
```python
df_no_nulls = df[df[coll].notnull() & df[colr].notnull()] # we only care about rows where both values are not null
inconsistent = df_no_nulls[coll] != df_no_nulls[colr])
```

fiendish: You used pandas.merge instead of merge_wo_duplicates. Try again with the actual function.

tests/conftest.py
```diff
@@ -15,16 +15,16 @@
 TEST_ROOT_DIR = os.path.abspath(os.path.dirname(__file__))
 TEST_DATA_DIR = os.path.join(TEST_ROOT_DIR, 'data')
 TEST_INGEST_OUTPUT_DIR = os.path.join(TEST_DATA_DIR,
-                                      'test_study',
+                                      'simple_study',
```
znatty22: Hmm so what are we actually using `test_study` for now? Do we need it still

tests/test_app_settings.py
```diff
@@ -17,7 +17,7 @@
     make_ingest_pipeline
 )
 
-TEST_STUDY_DIR = os.path.join(TEST_DATA_DIR, 'test_study')
+TEST_STUDY_DIR = os.path.join(TEST_DATA_DIR, 'simple_study')
```
znatty22: Is this all because `test_study` throws exceptions due to inconsistent data in merge results?

fiendish: yes

tests/conftest.py
fiendish: yes and yes

kf_lib_data_ingest/common/pandas_utils.py
znatty22: oh sorry, i thought i hit Start a Review but I had hit Add single comment. I went back and removed my previous comment bc i realized that i wasn't running your branch. I see that your code does work

PR #344, opened 2019-06-21
✨ Optional multithreaded loading

### LINKED ISSUE #34: Design and implement asynchronous loader

fiendish:
All entities within a group are thrown into a background worker pool unless the resume target hasn't been found yet.

This does not cross entity type group boundaries. That would require
significant enough extra complexity to ensure proper foreign key load
ordering that I don't think the minimal extra gain would be worth it.

Changes in cli.py are just to add the existing keywords to `kidsfirst test` as well.

closes #34

### DIFF
--- a/kf_lib_data_ingest/etl/load/load.py
+++ b/kf_lib_data_ingest/etl/load/load.py
@@ -6,7 +6,7 @@ import logging
 import os
 from collections import defaultdict
 from pprint import pformat
-from threading import Lock
+from threading import Lock, current_thread, main_thread
 from urllib.parse import urlparse
 
 import pandas
@@ -314,6 +314,8 @@ class LoadStage(IngestStage):
         :param links: map between entity keys and foreign key source unique IDs
         :type links: dict
         """
+        if current_thread() is not main_thread():
+            current_thread().name = f'{entity_type} {entity_id}'
 
         # populate target uid
         target_id_value = body.get(self.target_id_key)
@@ -373,13 +375,13 @@ class LoadStage(IngestStage):
 
             if tgt_id:
                 req_method = 'PATCH'
-                id_str = f' {{{self.target_id_key}: {tgt_id}}}'
+                id_str = f'{{{self.target_id_key}: {tgt_id}}}'
             else:
                 req_method = 'POST'
-                id_str = f' ({entity_id})'
+                id_str = f'({entity_id})'
 
             self.logger.debug(f'Request body preview:\n{pformat(body)}')
-            log_msg = f'DRY RUN - {req_method} {endpoint}{id_str}'
+            done_msg = f'DRY RUN - {req_method} {endpoint} {id_str}'
         else:
             # send to the target service
             entity = self._submit(entity_id, entity_type, endpoint, body)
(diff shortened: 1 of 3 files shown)


kf_lib_data_ingest/etl/load/load.py
```diff
@@ -456,19 +458,34 @@ def _run(self, target_entities):
...
+                verb = 'Queueing' if self.use_async else 'Loading'
                 self.logger.info(
-                    f'Loading {entity_type} {entity_id} (#{i+1} of {total})'
-                )
-                self._load_entity(
-                    entity_type, endpoint, entity_id, body, links
+                    f'{verb} {entity_type} {entity_id} (#{i+1} of {total})'
                 )
 
+                if self.use_async and not self.resume_from:
```
znatty22: Maybe you could do something like this so you don't have to check `use_async` twice and call shutdown:

```python
if self.use_async and not self.resume_from:
    with ThreadPoolExecutor() as ex:
        ex.submit(
            self._load_entity,
            entity_type, endpoint, entity_id, body, links
        )
```

kf_lib_data_ingest/etl/load/load.py
```diff
@@ -456,19 +458,34 @@ def _run(self, target_entities):
...
                 body = entity['properties']
                 links = entity['links']
 
+                verb = 'Queueing' if self.use_async else 'Loading'
                 self.logger.info(
-                    f'Loading {entity_type} {entity_id} (#{i+1} of {total})'
-                )
-                self._load_entity(
-                    entity_type, endpoint, entity_id, body, links
+                    f'{verb} {entity_type} {entity_id} (#{i+1} of {total})'
```
znatty22: Hmm now that we're loading async, seeing the current count of total in this log message is not as useful bc the Queueing log messages appear in batches in the log. 

Could we modify the log message on https://github.com/kids-first/kf-lib-data-ingest/blob/d0071ea0049c089bcc9d53c8f6203adc668c6074/kf_lib_data_ingest/etl/load/load.py#L389 so that it reports the current load count of the total?

So this:
```
Loaded participant 30 --> {kf_id: PT_EGE893F9}
```

would become: 
```
Loaded participant 30 --> {kf_id: PT_EGE893F9} (#605 of 2310)
```

When the log messages are scrolling by super fast, you're usually seeing those ^ log messages so seeing the load count out of the total gives you a good idea of progress.

kf_lib_data_ingest/etl/load/load.py
fiendish: You need the creation of the pool to happen once at the start. If you create one with a context manager it automatically calls shutdown as soon as the context ends, so what you have there would not actually run anything in parallel.

I can remove lines 462 and 486 if you want, which would make it always create a pool and then always shut it down regardless of whether anything is going to be added to it. 

I can also put the whole block from line 462 until 487 inside of a context manager, but then that would be an additional nest indenting everything in order to save one line calling shutdown.

kf_lib_data_ingest/etl/load/load.py
fiendish: √

kf_lib_data_ingest/etl/load/load.py
znatty22: Oh woops yea, my bad. I think its fine as is

kf_lib_data_ingest/etl/load/load.py
fiendish: https://github.com/kids-first/kf-lib-data-ingest/pull/344/commits/7538d86af1466ac180b49ccf5503d3448ea22c22

fiendish:
> I think the logs still look a little weird bc things happen out of order, but I'm ok with it for now. 👍

Want me to add the thread name to logger messages?

PR #347, opened 2019-06-21
✨ Add ETag to constants

### LINKED ISSUE #345: Add ETAG constant to list of HASHES
\**Should be high priority since its a blocker for loading genomic files* 

Need to do this so that we can load genomic files. Right now we cannot bc indexd rejects our current gf payload bc it uses `s3 etag`

znatty22:
Closes #345

### DIFF
--- a/kf_lib_data_ingest/common/constants.py
+++ b/kf_lib_data_ingest/common/constants.py
@@ -21,7 +21,7 @@ class FILE:
         SHA1 = 'SHA-1'
         SHA256 = 'SHA-256'
         SHA512 = 'SHA-512'
-        S3_ETAG = 'S3 ETag'
+        S3_ETAG = 'ETag'
 
 
 class SPECIMEN:


znatty22:
@fiendish 

> we shouldn't have one that doesn't work

The concept schema doesn't care if a constant "works" or doesn't work, bc it doesn't know how the constants are used. 

> being able to change the value in one location is also one of the reasons why we use the constants instead of raw strings in the first place.

True. Although naming it S3_TAG and making its value `etag` is a little 🤔bc I'd be wondering what's the S3 part of this? Anyway, I can do as you suggested bc it probably doesn't matter all that much. We can change it in the future if necessary.

fiendish:
> The concept schema doesn't care if a constant "works" or doesn't work

The schema doesn't, but we know now that indexd will do the wrong thing.

> I'd be wondering what's the S3 part of this?

The only ETag format that indexd accepts is specifically the S3 format.

PR #349, opened 2019-06-25
👷 Updates Constants

liberaliscomputing:
This PR adds/updates constants across multiple entities.

### DIFF
--- a/kf_lib_data_ingest/common/constants.py
+++ b/kf_lib_data_ingest/common/constants.py
@@ -5,14 +5,16 @@ Use preset constants instead.
 
 
 class COMMON:
+    CANNOT_COLLECT = 'Not Allowed To Collect'
+    MULTIPLE = 'Multiple'
     NOT_APPLICABLE = 'Not Applicable'
     NOT_AVAILABLE = 'Not Available'
     NOT_REPORTED = 'Not Reported'
-    UNKNOWN = 'Reported Unknown'
-    CANNOT_COLLECT = 'Not Allowed To Collect'
     NO_MATCH = 'No Match'
     OTHER = 'Other'
-    MULTIPLE = "Multiple"
+    UNKNOWN = 'Reported Unknown'
+    TRUE = True
+    FALSE = False
 
 
 class FILE:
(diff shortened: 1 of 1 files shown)


kf_lib_data_ingest/common/constants.py
```diff
@@ -26,49 +28,75 @@ class HASH:
...
+        XENOGRAFT = 'Xenograft Tissue'
 
-    class TISSUE_TYPE:
-        TUMOR = 'Tumor'
-        GERMLINE = 'Normal'
+    class SAMPLE_PROCUREMENT:
+        AUTOPSY = 'Autopsy'
+        BIOPSY = 'Biopsy'
+        BLOOD_DRAW = 'Blood Draw'
+        SUBTOTAL_RESECTIONS = 'Subtotal Resections'
```
fiendish: Why is this plural?

kf_lib_data_ingest/common/constants.py
```diff
@@ -26,49 +28,75 @@ class HASH:
...
+        TOTAL_RESECTIONS = 'Gross Total Resections'
 
     class ANATOMY_SITE:
+        BONE_MARROW = 'Bone Marrow'
+        HAIR = 'Hair'
         SKULL = 'Skull'
+        NERVOUS = 'Central Nervous System'
```
fiendish:
```suggestion
        CNS = 'Central Nervous System'
```

kf_lib_data_ingest/common/constants.py
```diff
@@ -78,53 +106,106 @@ class REFERENCE_GENOME:
         HG19 = 'hg19'
 
     class PLATFORM:
+        GENOMICS = 'Complete Genomics'
         ILLUMINA = 'Illumina'
+        ION_TORRENT = 'Ion Torrent'
+        LS454 = 'LS454'
+        PACBIO = 'PacBio'
+        SOLID = 'SOLiD'
```
fiendish: What are these new platforms from?

kf_lib_data_ingest/common/constants.py
```diff
@@ -78,53 +106,106 @@ class REFERENCE_GENOME:
...
         class ST_JUDE:
             NAME = 'St Jude'
             KF_ID = 'SC_1K3QGW4V'
 
+        class TGEN:
+            NAME = 'The Translational Genomics Research Institute'
+            KF_ID = 'SC_KQ9JZG3P'
+
+        class VIRTUAL:
+            NAME = 'Virtual'
```
fiendish: What is this?

kf_lib_data_ingest/common/constants.py
```diff
@@ -78,53 +106,106 @@ class REFERENCE_GENOME:
...
+        LINKED_WGS = 'Linked-Read WGS (10x Chromium)'
+        MRNA = 'miRNA-Seq'
+        RNA = 'RNA-Seq'
         WGS = 'WGS'
+        WXS = 'WXS'
 
     class ANALYTE:
         DNA = 'DNA'
         RNA = 'RNA'
+        VIRTUAL = 'Virtual'
```
fiendish: What is this?

kf_lib_data_ingest/common/constants.py
liberaliscomputing: I believe these enums were added against GDC's data dictionary. On our API doc, subtotal resection and gross total resection are plural (https://kf-api-dataservice.kidsfirstdrc.org/#tag/Biospecimen%2Fpaths%2F~1biospecimens%2Fpost), but on GDC, they are singular (https://docs.gdc.cancer.gov/Data_Dictionary/viewer/#?view=table-definition-view&id=sample).

kf_lib_data_ingest/common/constants.py
liberaliscomputing: Thanks for the suggestion. I make that change.

kf_lib_data_ingest/common/constants.py
liberaliscomputing: They are from DS (https://kf-api-dataservice.kidsfirstdrc.org/#tag/SequencingExperiment%2Fpaths%2F~1sequencing-experiments%2Fpost) AND GDC (https://docs.gdc.cancer.gov/Data_Dictionary/viewer/#?view=table-definition-view&id=read_group&anchor=platform)

kf_lib_data_ingest/common/constants.py
```diff
@@ -78,53 +106,106 @@ class REFERENCE_GENOME:
...
 
 
 class STUDY:
     CANCER = 'Cancer'
     STRUCTURAL_DEFECT = 'Structural Birth Defect'
+    STRUCTURAL_DEFECT_AND_CANCER = 'Structural Birth Defect and Cancer'
 
     class STATUS:
```
znatty22: Why do we need anything besides PENDING? Will ingest lib ever change this status to anything besides PENDING?

kf_lib_data_ingest/common/constants.py
liberaliscomputing: For CBTTC, there are new data types:
```
HISTOLOGY_IMAGES = 'Histology Images'
OPERATION_REPORTS = 'Operation Reports'
PATHOLOGY_REPORTS = 'Pathology Reports'
RADIOLOGY_IMAGES = 'Radiology Images'
RADIOLOGY_REPORTS = 'Radiology Reports'
```

Since these are not actual genomic files, I had to create placeholding BSs with `analyte_type` `Virtual` which is required. In turn, the `VIRTUAL` SC is for these virtual samples since SC is required for BS.

kf_lib_data_ingest/common/constants.py
liberaliscomputing: As described above.

kf_lib_data_ingest/common/constants.py
liberaliscomputing: I tried to embrace possible use cases in the future since our data model has all those enums: https://github.com/kids-first/kf-api-dataservice/commit/dbdbec7af00cd9aef7ca1c01c117bb8787229b8a#diff-6623e14f4531dd961f714dda06c0faff

But, I've also used only `Pending` for my previous ingestion. If we only use `Pending` for ingestion and it never changes, is this field actually needed in DS?

kf_lib_data_ingest/common/constants.py
fiendish: Hmm. Singular would be more grammatically consistent.

fiendish:
I'm ok with all of these changes, except I'd like those plural resections to be singular.

kf_lib_data_ingest/common/constants.py
liberaliscomputing: Updated!

kf_lib_data_ingest/common/constants.py
znatty22: Yea, actually I don't think the ingest lib will every ingest a new study so we probably don't need any of these. @fiendish what do you think?

fiendish: I agree with you, but I also don't think it hurts to throw them in.

PR #357, opened 2019-07-19
🐛 Use relative path as extract output key instead of full path

### LINKED ISSUE #354: dict keys going into transform function should be relative to study dir, not full path
Using the full path makes it so that the keys can't be referenced during transform without either string processing or knowing specific environmental conditions that can't be known during configuration. Since all of the extract configs are in the extract config directory, there's no real reason to include path components outside of that.

fiendish:
Closes #354

### DIFF
--- a/kf_lib_data_ingest/etl/extract/extract.py
+++ b/kf_lib_data_ingest/etl/extract/extract.py
@@ -14,7 +14,12 @@ from kf_lib_data_ingest.common.file_retriever import (
     FileRetriever,
     split_protocol
 )
-from kf_lib_data_ingest.common.misc import clean_up_df, read_json, write_json
+from kf_lib_data_ingest.common.misc import (
+    clean_up_df,
+    clean_walk,
+    read_json,
+    write_json
+)
 from kf_lib_data_ingest.common.pandas_utils import split_df_rows_on_splits
 from kf_lib_data_ingest.common.stage import IngestStage
 from kf_lib_data_ingest.common.type_safety import (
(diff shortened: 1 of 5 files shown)


kf_lib_data_ingest/etl/extract/extract.py
```diff
@@ -37,17 +37,38 @@ def lcm(number_list):
...
-        if isinstance(extract_config_paths, list):
-            self.extract_configs = [ExtractConfig(config_filepath)
-                                    for config_filepath
-                                    in extract_config_paths]
-        elif isinstance(extract_config_paths, str):
-            self.extract_configs = [ExtractConfig(extract_config_paths)]
+
+        assert_safe_type(extract_config_dir, str)
+        self.extract_config_dir = extract_config_dir
+        self.extract_configs = [
```
znatty22: Can't you just do this? It seems to work
```python
        self.extract_configs = []
        for root, directories, filenames in os.walk(extract_config_dir):
            for filename in filenames:
                if os.path.splitext(filename)[-1] == '.py':
                    ec = ExtractConfig(
                        os.path.join(root, filename)
                    )
                    ec.config_file_relpath = os.path.relpath(
                        ec.config_filepath,
                        start=self.extract_config_dir
                    )
                    self.extract_configs.append(ec)
```
Then you don't have to iterate multiple times either

fiendish:
I moved clean_walk into common.misc, because it seems handy

kf_lib_data_ingest/etl/extract/extract.py
fiendish: this won't avoid hidden nasties

PR #359, opened 2019-07-19
🐛 Make sure the interim output gets written

### LINKED ISSUE #356: if transform fails to create unique keys, it doesn't write interim output
```
ERROR - No unique keys were created for table! There must be at least 1 unique key column in a table. Zero unique keys in a table means there is no way to identify any concept instances. Source of error is /study/output/GuidedTransformStage/transform_function_output/default.tsv
```

That file hasn't been written yet.

fiendish:
Closes #356

### DIFF
--- a/kf_lib_data_ingest/etl/transform/transform.py
+++ b/kf_lib_data_ingest/etl/transform/transform.py
@@ -415,9 +415,9 @@ class TransformStage(IngestStage):
         columns added to each DataFrame
         """
         self.logger.info('Begin unique key creation for standard concepts ...')
-        for extract_config_url, (source_file_url, df) in df_dict.items():
+        for reference, (df_name, df) in df_dict.items():
             # Insert unique key columns
-            self.logger.info(f'Creating unique keys for {source_file_url}...')
+            self.logger.info(f'Creating unique keys for {df_name}...')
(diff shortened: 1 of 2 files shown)


kf_lib_data_ingest/etl/transform/transform.py
```diff
@@ -415,9 +415,9 @@ def _insert_unique_keys(
         columns added to each DataFrame
         """
         self.logger.info('Begin unique key creation for standard concepts ...')
-        for extract_config_url, (source_file_url, df) in df_dict.items():
+        for reference, (df_name, df) in df_dict.items():
```
znatty22: I understand why you're renaming extract_config_url to reference (bc of #357) , but what is df_name? Isn't this still source_file_url?

fiendish: > Isn't this still source_file_url?

Nope. "Transform Module Output: default df"

znatty22: Oh I see, this is for the intermediate data so there is no source file url 👍

fiendish: Screen Shot 2019-07-19 at 10 43 48 AM

PR #364, opened 2019-07-24
♻️ Move request logging into session object

fiendish:


fiendish:
I don't want to use QA because then we'll be tracking the QA code level. :(

fiendish:
Bug should be fixed

PR #377, opened 2019-08-14
♻️ Factor out pandas read file functions

### LINKED ISSUE #367: Fix extract read_output bug
If ingest is run with a subset of stages (e.g. t), it reads in the previous stages output from disk. When pandas.read_csv is used to read in the output, it rounds floats in a weird way. This later causes failures in accounting when comparing extract stage and transform stage uniques:

See: 
image

This can be fixed by supplying pandas.read_csv with`float_precision='round_trip'`. See 
https://stackoverflow.com/questions/47368296/pandas-read-csv-file-with-float-values-results-in-weird-rounding-and-decimal-dig or by setting `dtype=object`

To implement this fix, we should ensure that all calls to pandas.read_csv use this keyword arg. To do this in a DRY way, maybe we can factor out the pandas read from disk functionality in extract stage's `_source_to_df` function (call it `pandas_safe_read`?). Then this function can be used anywhere in the code.  See #368

### LINKED ISSUE #368: Create pandas_safe_read function
Factor out the pandas read from disk functionality in extract stage's _source_to_df function.

fiendish:
Logic that was in the extract stage is now in datafile_readers, and the
readers are now smarter about binary vs ascii file modes.

closes #367 
closes #368

### DIFF
--- a/kf_lib_data_ingest/common/datafile_readers.py
+++ b/kf_lib_data_ingest/common/datafile_readers.py
@@ -2,6 +2,7 @@
 Contains file readers (file -> dataframe) for file types that benefit from
 extra help.
 """
+import io
 import os
 
 import pandas
@@ -9,8 +10,16 @@ import xlrd
 from pandas.api.types import is_file_like
 
 
+def __ascii_wrap(potentially_binary_buffer):
+    try:
+        return io.TextIOWrapper(potentially_binary_buffer)
+    except Exception:
+        return potentially_binary_buffer
+
+
 def read_excel_file(filepath_or_buffer, **kwargs):
-    """Return contents of an excel spreadsheet as a pandas DataFrame.
+    """
+    Return contents of an excel spreadsheet as a pandas DataFrame.
 
     :param filepath_or_buffer: an xls or xlsx spreadsheet file
     :type filepath_or_buffer: string (path) or file-like object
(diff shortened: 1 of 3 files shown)


kf_lib_data_ingest/common/datafile_readers.py
```diff
@@ -2,15 +2,24 @@
...
 """
+import io
 import os
 
 import pandas
 import xlrd
 from pandas.api.types import is_file_like
 
 
+def __ascii_wrap(potentially_binary_buffer):
```
znatty22: So why do you need this? I thought `pandas.read_csv` can read binary streams already

kf_lib_data_ingest/common/datafile_readers.py
```diff
@@ -32,5 +41,38 @@ def read_excel_file(filepath_or_buffer, **kwargs):
             )
 
     kwargs['engine'] = 'xlrd'
-    kwargs['dtype'] = object
+    kwargs['dtype'] = str
```
znatty22: Why did you change from object to str?

kf_lib_data_ingest/common/datafile_readers.py
fiendish: It can't with engine='python'.

kf_lib_data_ingest/common/datafile_readers.py
fiendish: I think str is more correct than object. Also Extract was attempting to set dtype to str in a commit made more recently than that line in datafile_readers.

kf_lib_data_ingest/common/datafile_readers.py
fiendish: See also https://github.com/pandas-dev/pandas/issues/23779

znatty22: I see 👍

PR #379, opened 2019-08-16
♻️💥 Change "load" to "read"

fiendish:
This removes the terminology collision between "load/read" of files and "load/send" to endpoints. It also moves the read function setting out of the read params setting. Existing ingest packages will need to be updated.


templates/my_study/extract_configs/extract_config.py
```diff
@@ -16,7 +16,10 @@
 source_data_url = 'file://../data/clinical.tsv'
 
 # TODO (Optional) Fill in special loading parameters here
-source_data_loading_parameters = {}
+source_data_read_params = {}
+
+# TODO (Optional) You can set a custom read function with
+# source_data_read_function
```
znatty22:
```suggestion
# source_data_read_func
```

PR #397, opened 2019-10-24
🔧 You can't observe differently at the same time

fiendish:
Yes/No isn't valuable to the uniqueness of a phenotype entity, since you
can only observe it to be one way at a time.


fiendish:
preserving phenotype kfids isn't important

znatty22:
I agree its not. I'm just saying we have to do either 1 or 2 so that we don't accidentally load duplicate phenotypes when we run our ingest packages

fiendish:
yeah but option 2 is way easier

PR #403, opened 2019-12-10
♻️ Move reader selector out of ExtractStage

fiendish:
You can care less about which reader function you need to use for which
file format when testing

### DIFF
--- a/kf_lib_data_ingest/common/datafile_readers.py
+++ b/kf_lib_data_ingest/common/datafile_readers.py
@@ -76,3 +76,38 @@ def read_json_file(filepath_or_buffer, **kwargs):
     """
     kwargs["convert_dates"] = False
     return pandas.read_json(__ascii_wrap(filepath_or_buffer), **kwargs)
+
+
+def read_file(filepath_or_buffer, original_name=None, **kwargs):
+    """
+    Return contents of a data file as a pandas DataFrame. Wraps more specific
+    readers.
+
+    :param filepath_or_buffer: a data file
+    :type filepath_or_buffer: string (path) or file-like object
+    :param **kwargs: See docs for `pandas IO tools`
+    :return: The structured contents of the file
+    :rtype: pandas.Dataframe
+    """
+    if original_name is None:
+        if isinstance(filepath_or_buffer, str):
+            original_name = filepath_or_buffer
+        else:
+            original_name = filepath_or_buffer.name
+
+    if original_name.endswith((".xlsx", ".xls")):
+        read_func = read_excel_file
+    elif original_name.endswith((".tsv", ".csv")):
+        read_func = read_table_file
+    elif original_name.endswith(".json"):
+        read_func = read_json_file
+    else:
+        raise Exception(
+            f"Could not determine appropriate reader for '{original_name}'."
+            " You may need to define a custom read function."
+        )
+
+    try:
+        return read_func(filepath_or_buffer, **kwargs)
+    except Exception as e:
+        raise Exception(f"In {read_func.__name__} : {str(e)}") from e
--- a/kf_lib_data_ingest/common/io.py
+++ b/kf_lib_data_ingest/common/io.py
@@ -1,12 +1,14 @@
 """
-Contains file readers (file -> dataframe) for file types that benefit from
-extra help.
+Contains file readers for file types that benefit from extra help.
 """
 import io
+import json
 import os
 
+import jsonpickle
 import pandas
 import xlrd
+import yaml
 from pandas.api.types import is_file_like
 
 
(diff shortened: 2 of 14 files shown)


kf_lib_data_ingest/common/datafile_readers.py
```diff
@@ -76,3 +76,38 @@ def read_json_file(filepath_or_buffer, **kwargs):
...
     kwargs["convert_dates"] = False
     return pandas.read_json(__ascii_wrap(filepath_or_buffer), **kwargs)
+
+
+def read_file(filepath_or_buffer, original_name=None, **kwargs):
+    """
+    Return contents of a data file as a pandas DataFrame. Wraps more specific
+    readers.
+
+    :param filepath_or_buffer: a data file
```
znatty22: Need docstring for `original_name`

kf_lib_data_ingest/common/datafile_readers.py
```diff
@@ -76,3 +76,38 @@ def read_json_file(filepath_or_buffer, **kwargs):
...
+    """
+    if original_name is None:
+        if isinstance(filepath_or_buffer, str):
+            original_name = filepath_or_buffer
+        else:
+            original_name = filepath_or_buffer.name
+
+    if original_name.endswith((".xlsx", ".xls")):
+        read_func = read_excel_file
+    elif original_name.endswith((".tsv", ".csv")):
```
znatty22: Should we add `.txt`? Excel uses this ext when exporting to tab separated value. Also GitHub doesn't allow `.tsv`in file attachments to descriptions of things (issues, PR, etc).

kf_lib_data_ingest/common/datafile_readers.py
```diff
@@ -76,3 +76,38 @@ def read_json_file(filepath_or_buffer, **kwargs):
     """
     kwargs["convert_dates"] = False
     return pandas.read_json(__ascii_wrap(filepath_or_buffer), **kwargs)
+
+
+def read_file(filepath_or_buffer, original_name=None, **kwargs):
```
znatty22: Maybe this module should be called `io` and we can put all of the read/write functions in here not just ones for data files. There are a few in `kf_lib_data_ingest.common.misc`

Could do in this PR or maybe for a separate PR?

kf_lib_data_ingest/common/datafile_readers.py
fiendish: > Should we add .txt?

No, because txt doesn't indicate any particular data layout so you can't convert it to a dataframe without some serious additional intervention.

fiendish: >  Also GitHub doesn't allow .tsvin file attachments to descriptions of things (issues, PR, etc).

That sounds like a dumb bug in github

kf_lib_data_ingest/common/datafile_readers.py
fiendish: Hm. Maybe. We'd have to change the function names, because these are all for returning dataframes. There's another read_json in misc for returning a python dict.

kf_lib_data_ingest/common/io.py
```diff
@@ -0,0 +1,169 @@
+"""
+Contains file readers (file -> dataframe) for file types that benefit from
```
znatty22:
```suggestion
Contains file readers for file types that benefit from
```
Not all of the methods are for reading into a DataFrame

kf_lib_data_ingest/common/io.py
```diff
@@ -0,0 +1,169 @@
...
+                filename=filepath_or_buffer, logfile=devnull
+            )
+
+    kwargs["engine"] = "xlrd"
+    kwargs["dtype"] = str
+    kwargs["na_filter"] = False
+    return pandas.read_excel(wb, **kwargs)
+
+
+def read_text_df(filepath_or_buffer, **kwargs):
```
znatty22: I liked `read_table_df` better. `read_text_df` makes it sound like you're reading .txt files or something

fiendish: You said you wanted to read txt files tho, lol

kf_lib_data_ingest/common/io.py
fiendish: √

kf_lib_data_ingest/common/io.py
fiendish: What about read_delimited_df?

znatty22: > You said you wanted to read txt files though, lol

Ohhh sorry I thought you decided not to add reading of .txt extension.

PR #404, opened 2019-12-13
✅ Keep FileRetriever().get() safe

fiendish:
Add a test to stop a future change to FileRetriever from breaking functionality that
allows to call FileRetriever().get() as input to a read function even
though the temporary downloads go away when the FR leaves scope.

### DIFF
--- a/kf_lib_data_ingest/network/oauth2.py
+++ b/kf_lib_data_ingest/network/oauth2.py
@@ -54,7 +54,7 @@ def get_service_token(provider_domain, audience, client_id, client_secret):
     if response.status_code != 200:
         logger.error(
             f"Could not fetch access token from {oauth_token_url}! "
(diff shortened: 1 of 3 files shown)


fiendish:
yes

znatty22:
Still kinda confused. Why don't you just call this PR something more relevant to the change - "Add missing FileRetriever test" or something. Also what's the future change? And why not add this test in that PR?

fiendish:
There is no future change. I'm preventing a future change.

fiendish:
An important side-effect of the way that the file retriever currently works is that you can use it to get data without worrying about the FR lifetime even though the FR automatically deletes its temp files as soon as it leaves scope. That needs to be preserved, which means there needs to be a test that enforces it. That's this.

fiendish:
Now if I can just get jenkins to agree!

Issue #408, opened 2020-01-07
Persist stage output to remote database in addition to local CSVs


fiendish:
* Configuration for database location/creds
* Persist intermediate storage in warehouse
  1. Drop extract and transform tables for the current study from the warehouse
  2. Load new extract and transform tables into the warehouse
      - Each study will have its own database
      - Inside a study's database, each stage's output tables will be persisted to its own schema

https://github.com/d3b-center/clinical-data-distribution/issues/2

znatty22:
cc @fiendish 

Including @liberaliscomputing suggestion here so we remember:

- Ability to turn on/off loading into warehouse via CLI param or ingest config param

liberaliscomputing:
I am moving #414 here as a comment:

The current warehouse persistence is done by:

- https://github.com/kids-first/kf-lib-data-ingest/blob/optimize-accounting/kf_lib_data_ingest/etl/extract/extract.py#L136-L142
- https://github.com/kids-first/kf-lib-data-ingest/blob/optimize-accounting/kf_lib_data_ingest/etl/transform/guided.py#L368-L374

PCGC's default transform output has more than 200K rows. Since `to_sql` defaults to inserting all rows one by one, the velocity of insertion is pretty slow. This behavior can be adjusted by _chunksize_ and _method_ args (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_sql.html) to achieve faster insertion.

fiendish:
> Since to_sql defaults to inserting all rows one by one

That's not what the documentation says. It says the default is to send them all _at once_.

PR #411, opened 2020-01-09
🔥 Burn down the graph code because we'll never use it as is.

fiendish:
Let's not keep unmaintained zombie code active in master.
For an archive of the code see the graph-merge code branch
( https://github.com/kids-first/kf-lib-data-ingest/tree/graph-merge )

This is clean up work in preparation for #408


kf_lib_data_ingest/target_apis/kids_first.py
```diff
@@ -1,5 +1,5 @@
 """
-Configuration module specifying how a target model maps to the standard model
+Configuration module specifying how a target model maps to the standard model.
```
znatty22: Not sure where to put this comment, but you can get rid of `relationships` at the bottom of this file. It was used in the graph merging code.

kf_lib_data_ingest/etl/configuration/target_api_config.py
```diff
@@ -362,10 +358,10 @@ def _validate_relationships(self):
         """
```
znatty22: You can get rid of this method since we're getting rid of `relationships` in `kids_first.py`.

fiendish:
I'll rebase after

fiendish:
rebased. jenkins bjilding.

PR #415, opened 2020-01-15
♻️ Refactor message packing out of transform

fiendish:


kf_lib_data_ingest/etl/load/load.py
```diff
@@ -9,22 +9,26 @@
 from threading import Lock, current_thread, main_thread
```
znatty22: Can you add a module docstring and briefly describe what loader does?

kf_lib_data_ingest/etl/load/message_packer.py
```diff
@@ -0,0 +1,714 @@
+import logging
```
znatty22: Can you add a module docstring to briefly describe module purpose

kf_lib_data_ingest/etl/load/message_packer.py
```diff
@@ -0,0 +1,714 @@
...
+
+class MessagePacker:
+    def __init__(
+        self, target_api_config, target_api_url=DEFAULT_TARGET_URL, logger=None
+    ):
+        self.logger = logger or logging.getLogger(type(self).__name__)
+        self.target_api_url = target_api_url
+        self.target_api_config = target_api_config
+
+    def pack_messages(self, df_dict):
```
znatty22: Add docstring plz

kf_lib_data_ingest/etl/load/message_packer.py
```diff
@@ -0,0 +1,714 @@
...
+
+                    self.logger.debug(
+                        f"Mapping {target_concept}.{attr} value {value} to "
+                        f"{mapped_value}"
+                    )
+                    conversions[attr] = mapped_value
+
+                    instance["properties"][attr] = mapped_value
+
+    # TODO: Eliminate these complex Unique Key generation functions.
```
znatty22: Maybe yes, but I need to understand this better. But yea, maybe in another PR we can refactor this

kf_lib_data_ingest/target_apis/kids_first.py
```diff
@@ -418,3 +418,100 @@ def validate():
                 f'Links schema for "{target_concept}" has an '
                 f"incorrect or null mapping, {link_key}: {tgt_id_mapping}"
             )
+
+
+# TODO: This should be merged into the target API config!
```
znatty22: What do you mean by that? This method should go in target api config?

fiendish: refactoring the target api config should make this additional section unnecessary

kf_lib_data_ingest/etl/load/load.py
```diff
@@ -1,5 +1,7 @@
 """
```
znatty22: Maybe for another PR but, do you think we should have an organization like this for load stage now? 

`kf_lib_data_ingest/etl/load/kids_first/load.py`

kf_lib_data_ingest/etl/load/message_packer.py
```diff
@@ -0,0 +1,741 @@
+"""
+The MessagePacker class turns dataframes of CONCEPT columns returned by the
```
znatty22:
```suggestion
The MessagePacker class turns Pandas DataFrames of CONCEPT columns returned by the
```

kf_lib_data_ingest/etl/load/message_packer.py
```diff
@@ -0,0 +1,741 @@
...
+        self.target_api_url = target_api_url
+        self.target_api_config = target_api_config
+
+    def pack_messages(self, df_dict):
+        """
+        Main entry method for building messages.
+
+        For each DataFrame in df_dict, convert the DataFrame containing all of
+        the mapped source data into a dict of lists of dicts representing
+        represent lists of target concept instances and convert null property
```
znatty22:
```suggestion
        lists of target concept instances. Convert null property
```

kf_lib_data_ingest/etl/load/message_packer.py
```diff
@@ -0,0 +1,741 @@
...
+        values in `target_instances` to acceptable values based on the type of
+        property as defined in target schema.
```
znatty22:
```suggestion
        property as defined in the target API config's schema.
```

kf_lib_data_ingest/etl/load/load.py
fiendish: no, why?

znatty22: Isn't this loader Kids First specific? And we will have a FHIR loader too right? So either rename the module or put it in its own package along with the message packer

fiendish: > Isn't this loader Kids First specific?

Not if the target API config gets refactored like I want

fiendish: It remains to be seen whether any of the actual loader code will need to be generalized

znatty22: Ok gotcha 👍

kf_lib_data_ingest/etl/load/load.py
```diff
@@ -417,53 +427,53 @@ def _load_entity(self, entity_type, endpoint, entity_id, body, links):
 
         # log action
         with count_lock:
+            msg_cache = os.path.join(self.stage_cache_dir, "SentMessages.txt")
```
znatty22: We already talked about this in planning but reiterating. Can we make this a JSON file and have each payload include the message body and URL

kf_lib_data_ingest/etl/load/load.py
```diff
@@ -534,3 +553,9 @@ def _run(self, target_entities):
                 f"⚠️ Could not find resume_from target '{self.resume_from}'! "
                 "Nothing was actually loaded into the target service."
             )
+
+        json_out = self.msg_out.replace(".txt", ".json")
```
znatty22: What happens if there is an exception in `_run` before writing out. Maybe we want something like:

```
def _run(blah):
    try:
        do _run stuff
    except Exception as e:
        write output
        raise e
    else:
        write output
```

fiendish: only if you want it to look like it completed

znatty22: I think we do. In fact, I think I'd look at that file especially if there was some sort of exception just to see what got sent before it failed

PR #417, opened 2020-01-24
🐛 Fix some missing/broken bits of warehousing

fiendish:
Gives ingest schemas a common prefix so that they can be properly removed and then also the package directory so that studies can have multiple packages.

### DIFF
--- a/kf_lib_data_ingest/app/cli.py
+++ b/kf_lib_data_ingest/app/cli.py
@@ -30,6 +30,59 @@ def common_args_options(func):
         type=click.Path(exists=True, file_okay=True, dir_okay=True),
     )(func)
 
+    # Disable warehousing
+    func = click.option(
+        "--no_warehouse",
+        default=False,
+        is_flag=True,
+        help="A flag to skip sending data to the warehouse db.",
+    )(func)
+
+    # Resume loading from
+    func = click.option(
+        "--resume_from",
+        default=None,
+        help=(
+            "Dry run until a given target ID, and then load starting " "there."
+        ),
+    )(func)
+
+    # Multithreaded loading
+    func = click.option(
+        "--use_async",
+        default=False,
+        is_flag=True,
+        help=(
+            "A flag specifying whether to use sync or async loading "
+            "into the target service"
+        ),
+    )(func)
+
+    # Stages
+    func = click.option(
+        "--stages",
+        "stages_to_run_str",
+        default=DEFAULT_STAGES_TO_RUN_STR,
+        show_default=True,
+        type=click.Choice(VALID_STAGES_TO_RUN_STRS),
+        help=(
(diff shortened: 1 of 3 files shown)


kf_lib_data_ingest/common/warehousing.py
```diff
@@ -1,6 +1,9 @@
 from urllib.parse import urlparse
 
 import sqlalchemy
+from sqlalchemy.schema import DropSchema
+
+SCHEMA_PREFIX = "ingest:"
```
znatty22: Nitpick, but can we make it `Ingest` since the rest of the schema will be camel cased. Then it'll look like `Ingest:ExtractStage`

fiendish:
> I'm surprised our unit tests involving the test command didn't fail 🤔

conftest.py disables warehousing

fiendish:
> I think it should be on of the CLI common options so that it can be used with test command too

You really want to warehouse stuff that isn't being loaded anywhere?

kf_lib_data_ingest/common/warehousing.py
fiendish: 🤷‍♂  ok

znatty22:
> You really want to warehouse stuff that isn't being loaded anywhere?

Yep. Especially because now I have to run through all of the stages to get the accounting results.  Also, CI for `kf-ingest-packages` runs the `test` command rather than the `ingest` command because we don't want to do a real load. We just want to make sure the package runs without exceptions and that it passes accounting checks (which we will later move to an automated process  which runs on the warehouse)

fiendish:
> Also, CI for kf-ingest-packages runs the test command rather than the ingest command because we don't want to do a real load.

CI for ingest packages should be launching a dummy dataservice and doing a real load.

znatty22:
> CI for ingest packages should probably be launching a dummy dataservice and doing a real load? Or at least something should.

I agree, but that's a separate issue. I would still like the ability to turn on/off warehousing for both real ingest and test ingest.

kf_lib_data_ingest/app/cli.py
```diff
@@ -30,44 +30,32 @@ def common_args_options(func):
         type=click.Path(exists=True, file_okay=True, dir_okay=True),
     )(func)
 
-    # App settings
+    # Disable warehousing
```
znatty22: I thought order of these mattered...I think this option has to be last? Since you have it as the last parameter in the method signature

fiendish: The order just affects how they show in the help. Parameters are passed to the command callbacks as keyword arguments. That's why if you remove any of them from the function signature you'll get e.g. `TypeError: test() got an unexpected keyword argument 'resume_from'`

fiendish:
squashed a commit and updated emoji

PR #420, opened 2020-01-30
📝 Make template transform_module validate

fiendish:
Part of https://github.com/d3b-center/clinical-data-flow/issues/22

The tutorial wastes space and causes confusion by delivering a
broken-by-default transform module which then causes a *different*
validation error than the one indicated in the testing how-to guide.
Let's just deliver a good one.


templates/my_study/transform_module.py
```diff
@@ -48,8 +49,14 @@ def transform_function(mapped_df_dict):
     }
 
     """
-
-    df = list(mapped_df_dict.values())[0]
+    df = mapped_df_dict['extract_config.py']
+
+    # df = outer_merge(
```
znatty22: Why are these lines commented out? Isn’t this what we want to make the template package functional. Or are we assuming someone will step through the tutorial and uncomment these lines? If the latter, then this needs to be made clearer in the tutorial

docs/source/tutorial/transform.rst
```diff
@@ -36,81 +36,12 @@ The ``transform_function`` method has only one argument, the ``mapped_df_dict``
...
-                         family_and_phenotype_df,
-                         on=CONCEPT.PARTICIPANT.ID,
-                         with_merge_detail_dfs=False)
+Modify the transform function so that it merges your extracted DataFrames
+appropriately according to the function docstring for transform_function above.
 
 There are a few important things to note in this step:
 
-1. We outer merged the DataFrames
+1. We outer merged DataFrames so that we don't lose data
```
znatty22: These pointers don’t make sense now because they’re talking about a section that was removed. 

Why are we removing the transform function code? I think the entire point of the tutorial is to show step by step how to develop a working ingest package. Both extract stage and transform stage tutorial sections should result in working code. Shouldn’t we still write out a working transform function here?

templates/my_study/transform_module.py
fiendish: > Why are these lines commented out?

Because they're not trash, but they also break the module if you try to test extract before getting to the transform instructions.

>  Isn’t this what we want to make the template package functional.

Only for the transform section, which is why it's still there as a comment.

docs/source/tutorial/transform.rst
fiendish: > Why are we removing the transform function code?

Because you can't say to test after the extract section and also require code that only works after the transform section.

PR #423, opened 2020-02-07
✨ Add enrollment age concept

chris-s-friedman:
added an enrollment age concept to handle when age there is an age at enrollment variable


fiendish:
please run 
```
black --line-length 80 kf_lib_data_ingest tests
flake8 --ignore=E501,W503,E203 kf_lib_data_ingest tests
```
before committing

chris-s-friedman:
sorry about that - I was confused because pep-8 and the [default for black](https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length) conflict in what line length should be and the developer handbook isn't clear (at least to me) on which has precedence.

PR #433, opened 2020-02-18
🐛 Don't wipe whole db for each package

### LINKED ISSUE #432: Only reset schemas for current package, not whole study
@znatty22 looks like you were right

https://github.com/kids-first/kf-lib-data-ingest/blob/3d524ced39af11d37b6a8a392c0fcbee66ae31d0/kf_lib_data_ingest/common/warehousing.py#L41-L44 should be more scalpel and less grenade

### LINKED ISSUE #437: Don't use slashes in warehouse schema names
Metabase doesn't do well with forward or backward slashes in schema names. Let's use colons, these seem to work.

@fiendish can you roll this into #433 and rename the PR

fiendish:
closes #432 closes #437


docs/source/design/warehousing.rst
```diff
@@ -67,7 +67,8 @@ one of that stage's returned DataFrames.
 
 .. image:: /_static/images/warehouse_inside_db.png
 
-.. note::
+Lifecycle
+=========
 
-    Details for the warehouse used for ingesting Kids First studies can be
```
znatty22: Can we leave the link to the warehouse readme? It has the connection details that are specific to KF. Maybe it also redundantly mentions how data is cleared out but I think its ok to be redundant.

fiendish: the https://github.com/d3b-center/clinical-data-flow/tree/master/warehouse#data-organization section there has wrong information

znatty22: Yea can you make a PR for it to correct it?

fiendish:
closes #437

PR #441, opened 2020-02-27
🐳 Rm --no-cache-dir from pip install

znatty22:
Hopefully will speed up Jenkins builds

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,9 @@ RUN apt-get update && apt-get install -y jq wget && \
 
 COPY . /app
 
-RUN pip install --upgrade pip && \
+RUN python3 -m venv venv && \
+    . venv/bin/activate && \
+    pip install --upgrade pip && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -11,7 +11,9 @@ RUN apt-get update && apt-get install -y jq wget && \
 
 COPY . /app
 
-RUN pip install --upgrade pip && \
+RUN python3 -m venv venv && \
+    . venv/bin/activate && \
+    pip install --upgrade pip && \
     pip install --no-cache-dir -e .
```
liberaliscomputing: Isn't this PR for removing --no-cache-dir from pip install?

znatty22: yea we were just testing stuff out bc we're seeing unexpected behavior on Jenkins

znatty22:
Closing this PR since it doesn't fix the undesired behavior on Jenkins. We need to dig into the issue further at a later time.

PR #455, opened 2020-06-04
🐛 Fix python packages 

### LINKED ISSUE #456: Fix kf-lib-data-ingest python package distribution
The `kf-lib-data-ingest` Python package is missing a bunch of `__init__.py` files in its subpackages, causing the top level package to not install correctly. 

We missed this before because we were always using `pip install -e .` (install as editable) rather than `pip install .` See https://stackoverflow.com/questions/57157343/why-do-packages-installed-with-pip-install-e-not-need-init-py for details. 

Additionally, we were not including necessary non-Python files in our distribution such as our template ingest package (`kf-lib-data-ingest/templates`). 

We need to fix both issues to make `kf-lib-data-ingest` install properly.

znatty22:
Closes #456


znatty22:
@fiendish I removed `tests` dir from flake8 because it was complaining about a few uses of ambiguous variables like `l`, and `w`, etc, which are in some of the for-loops in our tests. I don't think it hurts if we don't lint tests. If you feel strongly about linting tests, I can add that back in and resolve the linting errors.

kf_lib_data_ingest/templates/my_study/extract_configs/extract_config.py
```diff
@@ -6,9 +6,9 @@
 information on writing extract config files.
 """
 
-from kf_lib_data_ingest.common import constants
+from kf_lib_data_ingest.common import constants  # noqa F401
```
liberaliscomputing, line 9: Why do we want some lines not to be checked by the linter? It seem neither exceeding the character limit nor break linting rules?

kf_lib_data_ingest/templates/my_study/transform_module.py
```diff
@@ -8,11 +8,12 @@
 implementing transform_function.
 """
 
-from kf_lib_data_ingest.common.concept_schema import CONCEPT
+from kf_lib_data_ingest.common.concept_schema import CONCEPT  # noqa F401
```
liberaliscomputing, line 11: Why do we want some lines not to be checked by the linter? It seem neither exceeding the character limit nor break linting rules?

kf_lib_data_ingest/templates/my_study/extract_configs/extract_config.py
znatty22: F401 is about having unused imports. Since this file is part of the template ingest package, the unused imports are part of the template for the extract_config.py template.

PR #458, opened 2020-06-08
♻️ Refactor target entity construction into plugin API

### LINKED ISSUE #444: Refactor UK generation with modular message packing and sending
* Replace UK generation with drastically simpler on-send mechanism.
* Merge target service api and message packer.
* Modularize get/patch/post mechanism in loader according to service definition.

This should resolve #194, obsolete #179, and answer #341

https://github.com/d3b-center/clinical-data-flow/issues/17
https://github.com/d3b-center/clinical-data-flow/issues/22

### LINKED ISSUE #450: Extract+Transform stages cache None as "" but don't read back the reverse
_read_output not proper mirror of _write_output

### LINKED ISSUE #452: Add docs for load stage and service plugin architecture
- [x] What is service plugin, how to make one and use it
- [ ] Example of how to use just load stage to read in TSV and load into target service

fiendish:
closes #444
closes #450
closes #452


kf_lib_data_ingest/etl/configuration/log.py
```diff
@@ -10,18 +11,23 @@
     DEFAULT_LOG_OVERWRITE_OPT,
 )
 
-VERBOTEN_STRINGS = {
-    v: os.environ[v]
+VERBOTEN_PATTERNS = {
+    re.escape(os.environ[v]): f"<env['{v}']>"
     for k, v in SECRETS.__dict__.items()
     if not k.startswith("_") and v in os.environ and os.environ[v]
```
liberaliscomputing, line 17: Couldn't `v in os.environ and os.environ[v]` be `os.getenv('v')`?

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -0,0 +1,747 @@
...
+    target_id_concept = CONCEPT.PHENOTYPE.TARGET_SERVICE_ID
+    id_fields = {"kf_id", "participant_id"}
+
+    @staticmethod
+    def build_key(record):
+        assert None is not record[CONCEPT.PARTICIPANT.ID]
+        assert None is not record[CONCEPT.PHENOTYPE.NAME]
+        assert record[CONCEPT.PHENOTYPE.OBSERVED] in {
+            constants.PHENOTYPE.OBSERVED.NO,
+            constants.PHENOTYPE.OBSERVED.YES,
```
liberaliscomputing, line 200: The previous target api requires [`CONCEPT.PHENOTYPE.OBSERVED`](https://github.com/kids-first/kf-lib-data-ingest/blob/master/kf_lib_data_ingest/target_apis/kids_first.py#L481), not confining only to `'Positive'` or `'Negative'`. We may just do `assert None is not record[CONCEPT.PHENOTYPE.OBSERVED]`?

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -0,0 +1,747 @@
...
+    target_id_concept = CONCEPT.OUTCOME.TARGET_SERVICE_ID
+    id_fields = {"kf_id", "participant_id"}
+
+    @staticmethod
+    def build_key(record):
+        assert None is not record[CONCEPT.PARTICIPANT.ID]
+        assert record[CONCEPT.OUTCOME.VITAL_STATUS] in {
+            constants.OUTCOME.VITAL_STATUS.ALIVE,
+            constants.OUTCOME.VITAL_STATUS.DEAD,
+        }
```
liberaliscomputing, line 238: Same as above.

tests/data/bad_api.py
```diff
@@ -0,0 +1,62 @@
...
+    def transform_records_list(records_list):
+        pass
+
+    @staticmethod
+    def build_key(record):
+        pass
+
+    @staticmethod
+    def build_entity(record, key, get_target_id_from_record):
+        pass
```
liberaliscomputing, line 55: Should we keep this in `bad_api.py`? Isn't this a good template example?

fiendish: It's test data for testing bad apis

kf_lib_data_ingest/etl/configuration/log.py
fiendish:
```suggestion
    if not k.startswith("_") and os.environ.get(v)
```

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
fiendish: True, but I think we shouldn't actually load the phenotypes otherwise.

docs/source/design/load.rst
```diff
@@ -2,4 +2,13 @@
...
 
-- Target API config
+The Load stage feeds record data through a plugin interface that must be
+implemented for each target service.
+
+Target service plugins implement the necessary routines for converting record
+data coming in from the previous stage into message payloads and also
+implement the logic for sending those payloads to the target service.
+
+.. image:: /_static/images/load_stage.svg
```
znatty22, line 12: I think this is a really nice image which illustrates how the loader works but there are some key concepts that are a bit vague probably need some explanation. Can you add:

1. Something that describes what the unique key is, how its used, what the build key function is
2. Something that shows an example target entity (maybe even add an image of a JSON entity to your graphic for completeness)

Also, can you add the following examples (probably in How-To section):

1. Using the target API ingest plugin with the ingest pipeline (adding it to app settings, running `kidsfirst` with the custom app settings, etc.)

2.  Using the load stage as a standalone loader with a data file. 
     Basically exactly how we use it in the integration test? I'm thinking this is how most non-Kids First users will 
     use it (specifically thinking of NCPI FHIR WG participants who have their own data in csv/tsv files 
     somewhere).

kf_lib_data_ingest/etl/configuration/target_api_config.py
```diff
@@ -1,338 +1,196 @@
 """
-A class which first validates and then stores transformation configuration
```
znatty22, line 2: I think we need to update this concept of Target API Config every where in the code and docs to Target API Ingest Plugin or Target Ingest Plugin.

fiendish: whyyyyyyyy?

docs/source/design/load.rst
fiendish: > Using the load stage as a standalone loader with a data file.

I'm not sure exactly what you mean by this, but it sounds like something that I'd like to cover in a different PR after talking about it.

fiendish: done

docs/source/design/load.rst
```diff
@@ -1,5 +1,41 @@
...
+   constituent elements**. This is used internally for caching identifiers
+   assigned by the target service (received from the target service as part of
+   the response for sending an entity payload) so that they can be repeatably
+   assigned to the same entities again. (Target services often don't do this
+   themselves.)
+#. **Build a message payload that will be sent to the server to populate a new
+   (or updated) entity of its type**. The unique keystrings generated in the
+   previous steps are passed to this method as well so that they can be
+   included as part of the payload if desired.
+
```
znatty22, line 40: Can you add the following section to this page? I think the ID cache and unique keys are most easily understood with an example.

```
A Note on the ID Cache
----------------------

It may be easier to understand the purpose of the ID cache and
unique keys with an example:

Given the record:

+----------------+----------------+--------------------------+--------------------+
| PARTICIPANT.ID | PHENOTYPE.NAME | PHENOTYPE.EVENT_AGE_DAYS | PHENOTYPE.OBSERVED |
+================+================+==========================+====================+
|      P1        |  Cleft palate  |          765             |       Positive     |
+----------------+----------------+--------------------------+--------------------+

The unique keystring for a phenotype observation could be:
`p1-cleft-palate-765`.

And after submission to the target service, the returned identifier
for the phenotype observation entity might be something like: `PH_00001111`.

The ingest library will create a record in the ID cache which associates
this unique key with its target identifier like this:

+---------------------+---------------+
|            UID      |      TID      |
+=====================+===============+
| p1-cleft-palate-765 |  PH_00001111  |
+---------------------+---------------+

Now whenever the ingest library sees a record in the source data with the
unique key components above, it will always associate the resulting
phenotype observation with the unique keystring `p1-cleft-palate-765`
which will always point to the phenotype observation entity in the
target service identified by `PH_00001111`.
```

fiendish: added

PR #463, opened 2020-06-19
👽 Use token auth for prod study creator

### LINKED ISSUE #461: Switch production config to use token auth
Study creator is no longer supporting OAuth2 service auth via client credentials flow. We need to switch the production auth config to use token auth. This is how development mode works as well.

fiendish:
The study creator is moving away from oath2.

closes #461

part of d3b-center/clinical-data-flow#102

### DIFF
--- a/kf_lib_data_ingest/app/settings/development.py
+++ b/kf_lib_data_ingest/app/settings/development.py
@@ -4,19 +4,26 @@ Development settings for Kids First Ingest App
 import os
 
 from kf_lib_data_ingest.app.settings.base import (
-    AUTH_CONFIGS,
     SECRETS,
     TARGET_API_CONFIG,
 )
 
 TARGET_API_CONFIG = TARGET_API_CONFIG
 
-AUTH_CONFIGS.update(
-    {
-        "https://kf-study-creator.kidsfirstdrc.org/download/study": {
-            "type": "token",
-            "token_location": "header",
-            "token": os.environ.get(SECRETS.KF_STUDY_CREATOR_API_TOKEN),
-        }
-    }
-)
+AUTH_CONFIGS = {
+    "http://localhost:8080/download/study": {
+        "type": "token",
+        "token_location": "url",
+        "token": os.environ.get(SECRETS.KF_STUDY_CREATOR_API_TOKEN),
+    },
+    "http://kf-study-creator-dev.kids-first.io/download/study": {
+        "type": "token",
+        "token_location": "header",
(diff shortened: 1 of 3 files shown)


kf_lib_data_ingest/app/settings/development.py
```diff
@@ -4,19 +4,26 @@
...
+    "http://kf-study-creator-dev.kids-first.io/download/study": {
+        "type": "token",
+        "token_location": "header",
+        "token": os.environ.get(SECRETS.KF_STUDY_CREATOR_API_TOKEN),
+    },
+    "https://kf-study-creator.kidsfirstdrc.org/download/study": {
+        "type": "token",
+        "token_location": "header",
+        "token": os.environ.get(SECRETS.KF_STUDY_CREATOR_API_TOKEN),
+    },
```
liberaliscomputing, line 28: Isn't this for production?

fiendish: isn't what for production?

liberaliscomputing: Yeah, I see that you've separated the previous base part into `development.py` and `production.py`. My confusion is that while `development.py` has these:

```python
    "http://kf-study-creator-dev.kids-first.io/download/study": {
        "type": "token",
        "token_location": "header",
        "token": os.environ.get(SECRETS.KF_STUDY_CREATOR_API_TOKEN),
    },
    "https://kf-study-creator.kidsfirstdrc.org/download/study": {
        "type": "token",
        "token_location": "header",
        "token": os.environ.get(SECRETS.KF_STUDY_CREATOR_API_TOKEN),
    },
```

the second auth config seems to belong production:

```python
    "https://kf-study-creator.kidsfirstdrc.org/download/study": {
        "type": "token",
        "token_location": "header",
        "token": os.environ.get(SECRETS.KF_STUDY_CREATOR_API_TOKEN),
    },
```
And, actually `production.py` has the above, so we may remove the above from `development.py`?

fiendish: Then you wouldn't be able to run locally against the study creator without specifying the production file, which is not the default. Production stays as a separate file so that we can switch back away from tokens whenever the study creator supports service auth again.

liberaliscomputing: Ah, it got clearer. I thought the deployment of the study creator also consisted of DEV, QA, and PRD.

PR #464, opened 2020-07-08
⬆️ Upgrade boto3 and pandas

fiendish:


fiendish:
Also switches docker pull from python:3 to amancevice/pandas:1.0.5

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3
+FROM blackdentech/kf-pandas
```
znatty22, line 1: Lol is this temporary? Also what's different with this image vs the standard Python 3 image

fiendish: bakes in pandas because that python:3 image takes a million years to compile pandas and then also breaks when Cython is missing

PR #486, opened 2020-09-02
✨ Define validation test input structure

### LINKED ISSUE #479: Define attribute validation test inputs
Need a standard structure that defines what the allowable values are for each attribute that will be validated.

For example, how will we represent this?
Patient gender can only take on values: Female, Male, Other

See https://github.com/kids-first/kf-lib-data-ingest/blob/master/kf_lib_data_ingest/common/constants.py
Also see Dataservice schema.py files for each endpoint: https://github.com/kids-first/kf-api-dataservice/blob/master/dataservice/api/biospecimen/schemas.py

gsantia:
Closes #479 

Wasn't sure exactly where to put the file. Created a `validation` folder for it. Added some constants to the `common.constants.py` file that appear in data tracker. This won't pass the CircleCI test because flake8 doesn't like lambda functions (`E731`).

### DIFF
--- a/kf_lib_data_ingest/validation/test_input_structure.py
+++ b/kf_lib_data_ingest/validation/test_input_structure.py
@@ -1,0 +1,128 @@
+from dateutil import parser
+
+from kf_lib_data_ingest.common import constants
+from kf_lib_data_ingest.common.concept_schema import CONCEPT
+
+
+MIN_AGE_DAYS = 0
+MAX_AGE_DAYS = 32872
+NULL_VALUES = {
+    constants.COMMON.NOT_REPORTED,
+    constants.COMMON.CANNOT_COLLECT,
+    constants.COMMON.NOT_APPLICABLE,
+    constants.COMMON.UNKNOWN,
+    constants.COMMON.NOT_AVAILABLE
+}
+
+check_bool = lambda x: x in {
+        constants.COMMON.TRUE,
+        constants.COMMON.FALSE
+    }
+
+check_age = lambda x: (int(x) >= MIN_AGE_DAYS) and (int(x) < MAX_AGE_DAYS)
+
+def check_date(x):
+    try:
+        value = parser.parse(str(value)) if value else None
+    except ValueError as e:
+        return False
+    return True
+
+
+check_positive = lambda x: int(x) > 0
+
+
+INPUT_VALIDATION = {
+    # PARTICIPANT
+    CONCEPT.PARTICIPANT.RACE: lambda x: x in {
+          constants.RACE.WHITE,
+          constants.RACE.NATIVE_AMERICAN,
+          constants.RACE.BLACK,
+          constants.RACE.ASIAN,
+          constants.RACE.PACIFIC,
+          constants.COMMON.OTHER,
+          constants.RACE.MULTIPLE
+    },
+    CONCEPT.PARTICIPANT.ENROLLMENT_AGE_DAYS: check_age,
+    CONCEPT.PARTICIPANT.IS_PROBAND: check_bool,
+    CONCEPT.PARTICIPANT.IS_AFFECTED_UNDER_STUDY: check_bool,
+    CONCEPT.PARTICIPANT.ETHNICITY: lambda x: x in {
+        constants.ETHNICITY.HISPANIC,
+        constants.ETHNICITY.NON_HISPANIC
+    },
+    CONCEPT.PARTICIPANT.GENDER: lambda x: x in {
+        constants.GENDER.MALE,
+        constants.GENDER.FEMALE,
+        constants.COMMON.OTHER
+    },
+    CONCEPT.PARTICIPANT.SPECIES: lambda x: x in {
+        constants.SPECIES.DOG,
+        constants.SPECIES.HUMAN
+    },
+    CONCEPT.PARTICIPANT.CONSENT_TYPE: lambda x: x in {
+        constants.CONSENT_TYPE.DS_OBDR_MDS,
+        constants.CONSENT_TYPE.DS_OBD_MDS,
+        constants.CONSENT_TYPE.DS_OC_PUB_MDS,
+        constants.CONSENT_TYPE.HMB_MDS,
+        constants.CONSENT_TYPE.HMB_IRB,
+        constants.CONSENT_TYPE.DS_CHD,
+        constants.CONSENT_TYPE.DS_CHD_IRB,
+        constants.CONSENT_TYPE.GRU
+    },
+
+    # BIOSPECIMEN
+    CONCEPT.BIOSPECIMEN.EVENT_AGE_DAYS: check_age,
+    CONCEPT.BIOSPECIMEN.ANALYTE: lambda x: x in {
+        constants.SEQUENCING.ANALYTE.DNA,
+        constants.SEQUENCING.ANALYTE.RNA,
+        constants.SEQUENCING.ANALYTE.VIRTUAL,
+        constants.COMMON.OTHER
+    },
+    CONCEPT.BIOSPECIMEN.CONCENTRATION_MG_PER_ML: check_positive,
+    CONCEPT.BIOSPECIMEN.SAMPLE_PROCUREMENT: lambda x: x in {
+        constants.SPECIMEN.SAMPLE_PROCUREMENT.AUTOPSY,
+        constants.SPECIMEN.SAMPLE_PROCUREMENT.BIOPSY,
+        constants.SPECIMEN.SAMPLE_PROCUREMENT.SUBTOTAL_RESECTION,
+        constants.SPECIMEN.SAMPLE_PROCUREMENT.TOTAL_RESECTION,
+        constants.SPECIMEN.SAMPLE_PROCUREMENT.BLOOD_DRAW,
+        constants.COMMON.OTHER
+    },
+    CONCEPT.BIOSPECIMEN.SHIPMENT_DATE: check_date,
+    CONCEPT.BIOSPECIMEN.VISIBLE: check_bool,
+    CONCEPT.BIOSPECIMEN.VOLUME_UL: check_positive,
+
+    # BIOSPECIMEN_DIAGNOSIS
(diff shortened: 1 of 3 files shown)


fiendish:
> This won't pass the CircleCI test because flake8 doesn't like lambda functions (E731).

lol, pep8 says not to assign lambdas to variables and to use def instead

fiendish:
I fixed the formatting, pep8 compliance, and some bugs (null_values instead of null_wrapper on https://github.com/kids-first/kf-lib-data-ingest/pull/486/commits/13d143af044af4c21767dd707ce4a6f2a0a979f1#diff-2be91bfd4b499880833c0c5a86b3c661L195 and not verifying int-safe inputs). 

I'm not sure how I feel about using dateutil.parser.parse though. It might be too flexible? Like you can do parse("1") and it treats that as totally valid.

fiendish (commented):
I'm not sure how I feel about using dateutil.parser.parse though. It might be too flexible? Like you can do parse("1") and it treats that as totally valid.

gsantia:
This is where I found the date parsing stuff in the dataservice code:
https://github.com/kids-first/kf-api-dataservice/blob/b4d8983fa3dfbf9097675ffd46c828bfb7163f20/dataservice/api/common/custom_fields.py#L78

fiendish:
Marshmallow has deserializers for rfc822 and iso8601.

For the first, they use https://docs.python.org/3/library/email.utils.html#email.utils.parsedate
For the latter they have a verbose custom implementation, but we could just use datetime.datetime.fromisoformat. (requires replacing trailing "Z" with "+00:00")

chris-s-friedman:
I don’t think it’s that high a priority. It’s relatively easy to see if something isn’t a date upon visual inspection of unique values.

fiendish (approved):

kf_lib_data_ingest/validation/test_input_structure.py
```diff
@@ -0,0 +1,230 @@
...
+        constants.STUDY.STATUS.WAITING,
+    },
+    # STUDY_FILE
+    CONCEPT.STUDY_FILE.AVAILABILITY: {
+        constants.GENOMIC_FILE.AVAILABILITY.IMMEDIATE,
+        constants.GENOMIC_FILE.AVAILABILITY.COLD_STORAGE,
+    },
+    # TASK
+    # TASK_GENOMIC_FILE
+}
```
znatty22, line 226: Maybe consider adding a validation function for enumerations so that every value in the `INPUT_VALIDATION` dict will have a standard calling interface. Right now the user of `INPUT_VALIDATION` needs to know whether the key has a set or function as its value: 

```
def check_enum_wrap(enum):
    def check_enum(x):
        return x in (enum | NULL_VALUES)
    return check_enum

INPUT_VALIDATION = {
    k: (check_enum_wrap(v) if isinstance(v, set) else v)
    for k, v in INPUT_VALIDATION.items()
}
```

kf_lib_data_ingest/validation/test_input_structure.py
```diff
@@ -0,0 +1,230 @@
+from dateutil.parser import parse as parsedate
```
znatty22, line 1: The name of this file doesn't seem to match what the content is doing. Can we rename this module to something else, maybe `value_validation.py` or could even do `values.py` since its within the `validation` package?

znatty22 (changes requested):

kf_lib_data_ingest/validation/test_input_structure.py
gsantia: Makes sense, no problem

kf_lib_data_ingest/validation/test_input_structure.py
```diff
@@ -0,0 +1,230 @@
...
+        constants.SEQUENCING.PLATFORM.ILLUMINA,
+        constants.SEQUENCING.PLATFORM.LS454,
+        constants.SEQUENCING.PLATFORM.GENOMICS,
+        constants.SEQUENCING.PLATFORM.PACBIO,
+        constants.COMMON.OTHER,
+    },
+    CONCEPT.SEQUENCING.TOTAL_READS: check_positive,
+    # SEQUENCING_EXPERIMENT_GENOMIC_FILE
+    # STUDY
+    CONCEPT.STUDY.RELEASE_STATUS: {
```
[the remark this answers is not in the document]
gsantia: 👍

znatty22:
@gsantia Can you please rebase this branch on `master-validation`? https://github.com/kids-first/kf-lib-data-ingest/tree/master-validation

Since validation is a big change/addition, we're going to merge all of our feature branches for validation changes into `master-validation` and then at the end we'll merge `master-validation` into `master`. This way anyone who is using master branch won't get affected by our validation work.

kf_lib_data_ingest/validation/values.py
```diff
@@ -0,0 +1,236 @@
...
+                except Exception:
+                    ret = False
+            return description, ret
+
+        return inner
+
+    return wrapper
+
+
+@try_wrap("Must be an integer x such that MIN_AGE_DAYS <= x <= MAX_AGE_DAYS")
```
znatty22, line 37:
```suggestion
@try_wrap(f"Must be an integer x such that {MIN_AGE_DAYS} <= x <= {MAX_AGE_DAYS}")
```
I think you want this to be an f string so that the actual vals for those min/max appear

kf_lib_data_ingest/validation/values.py
```diff
@@ -0,0 +1,236 @@
...
+NULL_VALUES = {
+    constants.COMMON.NOT_REPORTED,
+    constants.COMMON.CANNOT_COLLECT,
+    constants.COMMON.NOT_APPLICABLE,
+    constants.COMMON.UNKNOWN,
+    constants.COMMON.NOT_AVAILABLE,
+}
+
+
+def try_wrap(description):
```
znatty22, line 21: Can you add a doc string to this that describes the function briefly?

kf_lib_data_ingest/validation/values.py
```diff
@@ -0,0 +1,236 @@
...
+# Equates to 90 years
+MAX_AGE_DAYS = 32872
+NULL_VALUES = {
+    constants.COMMON.NOT_REPORTED,
+    constants.COMMON.CANNOT_COLLECT,
+    constants.COMMON.NOT_APPLICABLE,
+    constants.COMMON.UNKNOWN,
+    constants.COMMON.NOT_AVAILABLE,
+}
+
```
znatty22, line 19: Can you add a convenience function here which does validation for a given key and value. Something like:

```
def validate(key, value):
    try:
        desc, funct = INPUT_VALIDATION[key]
    except KeyError:
        raise KeyError(
            f'No validation function found for `{key}` in {__file__}.'
        )

    return funct(value)
```

znatty22 (changes requested):

kf_lib_data_ingest/validation/values.py
```diff
@@ -0,0 +1,236 @@
...
+    # TASK
+    # TASK_GENOMIC_FILE
+}
+
+
+INPUT_VALIDATION = {
+    k: (
+        (
+            f"Must be one of {(v | NULL_VALUES)}",
+            lambda x: x in (v | NULL_VALUES),
```
fiendish, line 230: Oh no. this isn't the same kind of function as the wrapped ones.

uhhhh....hmm.

What if we make it so that the innermost part of try_wrap does `ret = func(x)`, then change the wrapped functions to return bool instead of assert, and then wrap this lambda with `try_wrap(message)(thelambda)` instead of this tuple

fiendish (changes requested):

kf_lib_data_ingest/validation/values.py
fiendish: I'm opposed to this. Determination of which validations to run will be based on the contents of the INPUT_VALIDATION table anyway, so there will never be a key error.

kf_lib_data_ingest/validation/values.py
gsantia: yes oops

kf_lib_data_ingest/validation/values.py
gsantia: I'm not sure what to do here. Could use `INPUT_VALIDATION.get(key)` as a compromise?

znatty22: Oh I see so you're going to iterate over keys in `INPUT_VALIDATION` and run every function on the data?

fiendish: yes

fiendish:
I'm going to backtrack here. In practice the description wrapper is very clumsy because it means you don't get the description out until you call the function, but then also you call the function a million times so you get the description back a million times. Probably better to change from
```
@try_wrap("Must be an integer > 0")
 def check_positive(x):
     return int(x) > 0
```
to
```
 def check_positive(x):
     return int(x) > 0

described_check_positive = ("must be an integer > 0", check_positive)
```
and then assign check_positive_tuple instead of check_positive. Same for the sets.

Sorry.

fiendish:
So I did ^. One thing to discuss is that the tests said "must be int" but clearly the tests aren't validating that they must be int, so I changed to "must be number". Do we want the tests to test for intness?

gsantia:
> So I did ^.
> 
> One more thing to discuss is that the tests said "must be int" but clearly the tests aren't validating that they must be int, so I changed to "must be number". Do we want the tests to test for intness?

Some of the concepts don't need to be ints. Like `BIOSPECIMEN.CONCENTRATION_MG_PER_ML`

znatty22 (commented):
@gsantia @fiendish Changes look good, I'm ready to approve but the tests are failing. Once they pass, I can approve

gsantia:
> @gsantia @fiendish Changes look good, I'm ready to approve but the tests are failing. Once they pass, I can approve

Should pass now

znatty22 (approved):
✅ Looks good!

And I'm assuming we won't include unit tests for this module because we will have unit tests which test attribute value validation at a higher level.

PR #487, opened 2020-09-09
🐛 comment out bool json type cast

chris-s-friedman:
making a pr to get the ball rolling on the issue i found with booleans. 

When ingesting an ingest package, I found the setting `CONCEPT.GENOMIC_FILE.VISIBLE` to `constants.COMMON.FALSE` created the correct output in `sent_messages.json` but when inspecting in dataservice, visible was set to false.

Commenting out the 'boolean': bool line fixed this.

### DIFF
--- a/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
+++ b/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
@@ -646,7 +646,6 @@ all_targets = [
 
 swagger_cache = {}
 json_type_casts = {
-    "boolean": bool,
     "string": str,
     "integer": int,
     "number": float,


fiendish:
let's delete the line instead of commenting it out

fiendish:
@blackdenc why is this stuck on yellow? I thought we removed the jenkins link?

blackdenc:
> @blackdenc why is this stuck on yellow? I thought we removed the jenkins link?

I did, but it looks like it was a 'required' check, so it didn't remove it from the branch rule when I removed the webhook. I fixed it now, you should be good to go.

znatty22:
@fiendish @blackdenc Why are we removing the Jenkins status check? Jenkins does our integration test with the study creator API

blackdenc:
> @fiendish @blackdenc Why are we removing the Jenkins status check? Jenkins does our integration test with the study creator API

@znatty22 I re-wrote that as a github action to make it easier to migrate it to the kf-strides account. This should do everything that the Jenkins pipeline did without the overhead of having to manage a Jenkins pipeline.
https://github.com/kids-first/kf-lib-data-ingest/actions?query=workflow%3A%22Run+Tests%22

znatty22:
@blackdenc Ah that's right! And if the action fails, the PR will not be eligible for merging - is that correct?

blackdenc:
> @blackdenc Ah that's right! And if the action fails, the PR will not be eligible for merging - is that correct?

@znatty22 Right! It also won't build the container image, failing those checks is a full stop.

fiendish:
K. @chris-s-friedman plz squash the commits and then we'll merge

fiendish (approved):

fiendish (approved):

PR #492, opened 2020-09-14
🔥 Remove old accounting code

### LINKED ISSUE #489: Remove old accounting code
Remove obsolete accounting code since it will be replaced with calls to the new validation code.

znatty22:
Closes #489


fiendish:
What about user_defined_tests?

znatty22:
> What about user_defined_tests?

I think we should keep them. However 1 minor change is needed but I wanted to put that in a separate PR. The changes would be in the template ingest package's `conftest.py`. Instead of having fixtures for the concept_discovery_dicts, we can just have test input fixtures for extract and transform stage outputs. Then people can write custom tests given the full stage output.

fiendish (approved):
seems ok

PR #495, opened 2020-09-16
💬 Add concepts for participant template 

### LINKED ISSUE #494: Need new concepts for new templates
https://github.com/kids-first/kf-api-study-creator/blob/743e6ce82a14220666c4e4d26fc1faab0574cd0f/creator/analyses/file_types.py#L32-L40

* unitless ages and age units for everything, not just in days.
* age at last contact, not just at enrollment

fiendish:
Flexibly handle ages entered as split value/unit columns.
closes #494

### DIFF
--- a/kf_lib_data_ingest/common/concept_schema.py
+++ b/kf_lib_data_ingest/common/concept_schema.py
@@ -73,11 +73,12 @@ class CONCEPT:
         IS_AFFECTED_UNDER_STUDY = None
         SPECIES = None
         ENROLLMENT_AGE_DAYS = None
+        LAST_CONTACT_AGE_DAYS = None
 
         class ENROLLMENT_AGE(QuantityMixin):
             pass
 
-        class LAST_CONTACT(QuantityMixin):
+        class LAST_CONTACT_AGE(QuantityMixin):
             pass
 
     class OUTCOME(PropertyMixin):
--- a/kf_lib_data_ingest/common/concept_schema.py
+++ b/kf_lib_data_ingest/common/concept_schema.py
@@ -4,6 +4,11 @@ DELIMITER = "|"
 UNIQUE_ID_ATTR = "UNIQUE_KEY"
 
 
+class QuantityMixin(object):
+    VALUE = None
+    UNITS = None
+
+
 class FileMixin(object):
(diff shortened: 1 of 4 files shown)


gsantia (commented):
Everything looks good to me. Will we need additional validation tests to make sure that the concepts that inherit from `QuantityMixin` are fed numeric non-negative values into `VALUE`?

fiendish:
I suppose so

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -69,9 +74,19 @@ class PARTICIPANT(PropertyMixin):
         SPECIES = None
         ENROLLMENT_AGE_DAYS = None
 
+        class ENROLLMENT_AGE(QuantityMixin):
+            pass
+
+        class LAST_CONTACT(QuantityMixin):
```
znatty22, line 80: Why is this a quantity? Is it supposed to be an age as well?

kf_lib_data_ingest/validation/values.py
```diff
@@ -57,6 +55,12 @@ def inner(x):
     try_wrap(lambda x: int(x) >= 0),
 )
 
+age_units = {
```
znatty22, line 58: Kind of weird to have this here, can you move it to top of module?

znatty22 (changes requested):

kf_lib_data_ingest/common/concept_schema.py
fiendish: I think everything is supposed to be an age in KF?

kf_lib_data_ingest/validation/values.py
fiendish: Sure thing

fiendish: moved now

kf_lib_data_ingest/common/concept_schema.py
znatty22: Since all of our other age fields have the AGE suffix, should we add it to this one too?

znatty22 (approved):
👍

PR #502, opened 2020-09-21
✨ Integrate data validation into ingest pipeline

### LINKED ISSUE #481: Integrate validation into ingest + test CLI commands
Validation should be run automatically as a part of the test command. 

- Validation should run after both the extract and transform stages on each stage's output files
- There should also be a way to run the command with validation off
- A validation report for each stage should be written to the respective stage folder

For example, default behavior would be: 
```
$ kidsfirst test <ingest package dir> 
``` 
Would run:
1. Extract stage
2. Validation on Extract stage output
3. Transform stage
4. Validation on Transform stage output

And output:
`<ingest package dir>/output/ExtractStage/validaton_report.md`
`<ingest package dir>/output/TransformStage/validaton_report.md`

Run test command without validation:

```
$ kidsfirst test <ingest package dir> --no_validation
```

znatty22:
Closes #481

### DIFF
--- a/kf_lib_data_ingest/app/cli.py
+++ b/kf_lib_data_ingest/app/cli.py
@@ -10,6 +10,7 @@ import click
 
 from kf_lib_data_ingest.app import settings
 from kf_lib_data_ingest.config import DEFAULT_LOG_LEVEL, DEFAULT_TARGET_URL
+from kf_lib_data_ingest.common.stage import VALIDATION_MODES
 from kf_lib_data_ingest.etl.ingest_pipeline import (
     DEFAULT_STAGES_TO_RUN_STR,
     VALID_STAGES_TO_RUN_STRS,
@@ -18,6 +19,25 @@ from kf_lib_data_ingest.etl.ingest_pipeline import (
 
 CONTEXT_SETTINGS = {"help_option_names": ["-h", "--help"]}
 DEFAULT_LOG_LEVEL_NAME = logging._levelToName.get(DEFAULT_LOG_LEVEL)
+DEFAULT_VALIDATION_MODE = VALIDATION_MODES[1]
+VALIDATION_MODE_OPT = {
+    "args": ("--validation_mode",),
+    "kwargs": {
+        "default": DEFAULT_VALIDATION_MODE,
+        "type": click.Choice(VALIDATION_MODES),
+        "help": (
+            "Does not apply if --no_validate CLI flag is present. "
+            "The `basic` mode runs validation faster but may not find 100% of "
+            "the errors if the data contains implied connections between "
+            "entity types (e.g. You have a file that relates participants to "
+            "source genomic files and a file that relates participants to "
+            "biospecimens, instead of a file which relates participants "
+            "directly to biospecimens). The `advanced` mode runs validation "
+            "slower but is guarenteed to find all errors even if your data "
+            "has implied connections."
+        ),
+    },
+}
 
 
 def common_args_options(func):
@@ -114,6 +134,19 @@ def common_args_options(func):
         help=log_help_txt,
     )(func)
 
+    # Disable data validation
+    func = click.option(
+        "--no_validate",
+        default=False,
+        is_flag=True,
+        help="A flag to skip data validation during ingestion",
+    )(func)
+
+    # Validation mode
+    func = click.option(
+        *VALIDATION_MODE_OPT["args"], **VALIDATION_MODE_OPT["kwargs"]
+    )(func)
+
     return func
 
 
@@ -147,6 +180,8 @@ def ingest(
     dry_run,
     resume_from,
     no_warehouse,
+    no_validate,
+    validation_mode,
 ):
     """
     Run the Kids First data ingest pipeline.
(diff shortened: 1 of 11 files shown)


fiendish:
emoji on grapht-theory lib version should be :pushpin: ^_^

kf_lib_data_ingest/app/cli.py
```diff
@@ -18,6 +19,25 @@
 
 CONTEXT_SETTINGS = {"help_option_names": ["-h", "--help"]}
 DEFAULT_LOG_LEVEL_NAME = logging._levelToName.get(DEFAULT_LOG_LEVEL)
+DEFAULT_VALIDATION_MODE = VALIDATION_MODES[1]
```
fiendish, line 22: If the number of modes change, this will get weird. Can the modes be given their own constants and then use the right one directly?

kf_lib_data_ingest/app/cli.py
```diff
@@ -18,6 +19,25 @@
...
+        "help": (
+            "Does not apply if --no_validate CLI flag is present. "
+            "The `basic` mode runs validation faster but may not find 100% of "
+            "the errors if the data contains implied connections between "
+            "entity types (e.g. You have a file that relates participants to "
+            "source genomic files and a file that relates participants to "
+            "biospecimens, instead of a file which relates participants "
+            "directly to biospecimens). The `advanced` mode runs validation "
+            "slower but is guarenteed to find all errors even if your data "
+            "has implied connections."
```
fiendish, line 37: This description isn't right. We can't say whether advanced mode will flag more things or fewer, just that it will be more thorough. The relationships discovered by that thoroughness may be good or bad.

kf_lib_data_ingest/app/cli.py
```diff
@@ -234,12 +264,14 @@ def test(
...
     """
-
     # Make kwargs from options
     frame = inspect.currentframe()
     args, _, _, values = inspect.getargvalues(frame)
     kwargs = {arg: values[arg] for arg in args[1:]}
 
+    if kwargs.pop("no_validate"):
+        kwargs["validation_mode"] = None
+
```
fiendish, line 274: Not needed, because arguments are passed to def ingest which also does this.

kf_lib_data_ingest/app/cli.py
```diff
@@ -267,20 +299,43 @@ def create_new_ingest(dest_dir=None):
...
+    Validate files and write validation reports to
+    a subdirectory, `validation_results`, in the current working directory
+
+    \b
+    Arguments:
+        \b
+        file_or_dir - the path to the file or directory of files to validate
+    """
     from kf_lib_data_ingest.common.io import path_to_file_list
     from kf_lib_data_ingest.validation.validation import Validator
```
fiendish, line 314: Doesn't importing in the middle of a file violate pep8?

kf_lib_data_ingest/common/stage.py
```diff
@@ -159,19 +171,54 @@ def run(self, *args, **kwargs):
...
+        to source genomic files and a file that relates participants to
+        biospecimens, instead of a file which relates participants
+        directly to biospecimens).
 
-        :param df_dict: a dict of DataFrames returned by the _run() method
-        :return: a dict where concept values map to a list of the sources
-        containing them
-        :rtype: dict
+        The `advanced` mode runs validation slower but is guarenteed to find
+        all errors even if your data has implied connections.
```
fiendish, line 195: not accurate (as above)

kf_lib_data_ingest/validation/validation.py
```diff
@@ -26,53 +47,52 @@ def __init__(self, output_dir=None, setup_logger=True):
 
         if setup_logger:
```
fiendish, line 48: "setup" is a noun. "set up" is the verb form. See https://github.com/kids-first/kf-lib-data-ingest/pull/493/commits/d5f8a4fa416f2d3957cb3f70a5832c1d3aa07363

tests/test_ingest_pipeline.py
```diff
@@ -115,33 +117,31 @@ def test_ingest_cmd_missing_required_args():
...
+        (cli.test, [SIMPLE_STUDY_CONFIG, "--log_level", "debug"], 0),
         (
             cli.ingest,
             [SIMPLE_STUDY_CONFIG, "--dry_run", "--log_level", "debug"],
+            0,
         ),
     ],
 )
-def test_ingest_cmds(cli_cmd, arg_list):
+def test_ingest_cmds(cli_cmd, arg_list, expected_code):
```
fiendish, line 132: Did you plan to have one that doesn't exit 0?

kf_lib_data_ingest/app/cli.py
znatty22: Ok, can you give me a better description?

kf_lib_data_ingest/app/cli.py
znatty22: Not sure what you mean by this? You mean the ingest pipeline? CLI.ingest has two different options `no_validate` and `validation_mode`, and then the ingest pipeline only gets the `validation_mode` var.

kf_lib_data_ingest/app/cli.py
znatty22: The clean script didn't fix it..so no? We do this in other places in cli.py so if we change this one, then we should change all of them.

tests/test_ingest_pipeline.py
znatty22: Oh I think I changed this to take in an `expected_code` arg because before I had the ingest pipeline exit with code 1 when validation fails. Now the ingest pipeline will only exit with code 1 if something goes wrong in execution, so some sort of exception. So yea I could remove it since all exit codes are 0 right now

kf_lib_data_ingest/validation/validation.py
znatty22: Lol okie

PR #507, opened 2020-09-24
🐛 Fix path_to_file_list

### LINKED ISSUE #506: Path to file path list method doesn't prepend dir
The current method doesn't prepend the directory to the file name 

I think the method should be: 
```python
def path_to_file_list(file_or_dir, recursive=True):
    """
    Convert input which is either a file or a directory to a list of filepaths.

    :param file_or_dir: path to a file or directory
    :return: a list of filepaths
    """
    file_or_dir = os.path.abspath(os.path.expanduser(file_or_dir))
    files = []
    if os.path.isdir(file_or_dir):
        if recursive:
            for root, _, file_list in os.walk(file_or_dir):
                for file in file_list:
                    files.append(os.path.join(root, file))
        else:
            root, _, file_list = next(os.walk(file_or_dir))
            files.extend([os.path.join(root, f) for f in file_list])
    else:
        files.append(file_or_dir)
    return files
```

gsantia:
Modify `file_to_path_list` function so that the file paths include the root. Closes #506.

### DIFF
--- a/kf_lib_data_ingest/common/io.py
+++ b/kf_lib_data_ingest/common/io.py
@@ -181,12 +181,12 @@ def path_to_file_list(file_or_dir, recursive=True):
     files = []
     if os.path.isdir(file_or_dir):
         if recursive:
-            for _, _, file_list in os.walk(file_or_dir):
+            for root, _, file_list in os.walk(file_or_dir):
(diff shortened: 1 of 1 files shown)


znatty22 (approved):
✅

PR #509, opened 2020-10-05
🐛 Detect encodings of files when reading

### LINKED ISSUE #508: read_df confused by (useless in utf8) byte order marks
See: https://github.com/kids-first/kf-lib-data-ingest/pull/493#discussion_r497682553
And: https://docs.python.org/3/library/codecs.html#encodings-and-unicode

I think read_df should understand file encoding when reading files

fiendish:
closes #508

### DIFF
--- a/kf_lib_data_ingest/common/io.py
+++ b/kf_lib_data_ingest/common/io.py
@@ -10,11 +10,33 @@ import pandas
 import xlrd
 import yaml
 from pandas.api.types import is_file_like
+from cchardet import UniversalDetector
+from functools import partial
 
 
-def __ascii_wrap(potentially_binary_buffer):
+def __detect_encoding(potentially_binary_buffer):
+    def detect(buffer):
+        detector = UniversalDetector()
+        for chunk in iter(partial(buffer.read, 1024), b""):
+            detector.feed(chunk)
+            if detector.done:
+                break
+        detector.close()
+        buffer.seek(0)
+        return detector.result["encoding"]
+
     try:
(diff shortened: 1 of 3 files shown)


kf_lib_data_ingest/common/io.py
```diff
@@ -62,7 +74,11 @@ def read_delimited_text_df(filepath_or_buffer, **kwargs):
     kwargs["engine"] = "python"
     kwargs["dtype"] = str
     kwargs["na_filter"] = False
-    return pandas.read_csv(__ascii_wrap(filepath_or_buffer), **kwargs)
+    kwargs["encoding"] = kwargs.get("encoding") or __detect_encoding(
```
znatty22, line 77: Hmm is this going to slow things down in a noticeable way for big dfs? Couple options.. 1) Could make the default behavior what we had before (no supplied encoding to read_df) and then add an `auto_detect_encoding` kwarg in read_df or 2) Always assume large text, detect encoding incrementally, stop when you find it: https://chardet.readthedocs.io/en/latest/usage.html?#advanced-usage

znatty22 (changes requested):

kf_lib_data_ingest/common/io.py
fiendish: It shouldn't. It only reads a 2kb from the file.

znatty22: Ohhh ok I was looking at chardet. I'll approve

znatty22 (approved):
⚡

znatty22 (approved):
👍

PR #511, opened 2020-10-07
♻️ Auto-populate BIOSPECIMEN|ID, BIOSPECIMEN_GROUP|ID cols

### LINKED ISSUE #510: Auto-populate BIOSPECIMEN|ID, BIOSPECIMEN_GROUP|ID cols
We currently raise an exception in the transform stage if either of these columns are not populated. Instead, we should just populate one from the other if either one has data. If neither column exists, do nothing.

**To Do**

1. Remove [_validate_data](https://github.com/kids-first/kf-lib-data-ingest/blob/263d1bf798cfa473f16615ffbb97b0820e46b942/kf_lib_data_ingest/etl/transform/guided.py#L59) completely

2. Update [build_entity method](https://github.com/kids-first/kf-lib-data-ingest/blob/263d1bf798cfa473f16615ffbb97b0820e46b942/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py#L277) in KF dataservice target API config

gsantia:
Closes #510.

### DIFF
--- a/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
+++ b/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
@@ -266,11 +266,14 @@ class Biospecimen:
 
     @staticmethod
     def build_key(record):
-        assert None is not record[CONCEPT.BIOSPECIMEN_GROUP.ID]
+        biospecimen_group_id = record.get(
+            CONCEPT.BIOSPECIMEN_GROUP.ID, record.get(CONCEPT.BIOSPECIMEN.ID)
+        )
+        assert None is not biospecimen_group_id
         assert None is not record[CONCEPT.BIOSPECIMEN.ID]
         assert record[CONCEPT.SEQUENCING.CENTER.TARGET_SERVICE_ID]
         return record.get(CONCEPT.BIOSPECIMEN.UNIQUE_KEY) or join(
-            record[CONCEPT.BIOSPECIMEN_GROUP.ID], record[CONCEPT.BIOSPECIMEN.ID]
+            biospecimen_group_id, record[CONCEPT.BIOSPECIMEN.ID]
         )
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


znatty22 (approved):
👍

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -278,8 +278,12 @@ def build_entity(record, key, get_target_id_from_record):
...
             "kf_id": get_target_id_from_record(Biospecimen, record),
             "participant_id": get_target_id_from_record(Participant, record),
-            "external_sample_id": record.get(CONCEPT.BIOSPECIMEN_GROUP.ID),
-            "external_aliquot_id": record.get(CONCEPT.BIOSPECIMEN.ID),
+            "external_sample_id": record.get(
+                CONCEPT.BIOSPECIMEN_GROUP.ID, record.get(CONCEPT.BIOSPECIMEN.ID)
+            ),
+            "external_aliquot_id": record.get(
+                CONCEPT.BIOSPECIMEN.ID, record.get(CONCEPT.BIOSPECIMEN_GROUP.ID)
+            ),
```
fiendish, line 286: I can understand setting group ID from the sample if a group isn't designated, but should we really set the sample from group? I would think that explicitly setting group but not sample suggests that sample should have been set and is missing by accident?

znatty22: Hmm, yea I agree with @fiendish. If group is not set, use sample. If sample is not set, don't do anything

znatty22 (approved):
👌

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -278,7 +278,9 @@ def build_entity(record, key, get_target_id_from_record):
         return {
             "kf_id": get_target_id_from_record(Biospecimen, record),
             "participant_id": get_target_id_from_record(Participant, record),
-            "external_sample_id": record.get(CONCEPT.BIOSPECIMEN_GROUP.ID),
+            "external_sample_id": record.get(
```
fiendish, line 281: need to also do this for build_key above, otherwise the code will never get here

fiendish (changes requested):

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
gsantia: made some changes, is this what you meant?

fiendish: yup!

fiendish (approved):
☣️

PR #513, opened 2020-10-12
♻️ Break dependency on local ID cache

fiendish:
Relying entirely on the local cache makes it harder to move around and harder to recover. The target api configuration has been redone to make it possible to query the server for matching entities first. We can retool flags to prevent clearing the cache when needed.

This needs to be evaluated with the FHIR api as well before merging. https://github.com/kids-first/kf-model-fhir/pull/206


kf_lib_data_ingest/app/cli.py
```diff
@@ -57,21 +57,27 @@ def common_args_options(func):
         type=click.Path(exists=True, file_okay=True, dir_okay=True),
     )(func)
 
+    # Clear the UID cache
+    func = click.option(
+        "--clear_cache",
+        default=False,
+        is_flag=True,
+        help="Clear the identifier cache before loading into the target service.",
+    )(func)
```
liberaliscomputing, line 66: Say, I have two batches of data, one already ingested and the other new. 

- When set, does the ingest library remove an existing UID cache and send HTTP requests to the server to 1) GET the first batch's UIDs so it wouldn't create new resources and 2) POST the second batch resources?
- When unset, does the ingest library only send POST requests for the second batch resources?

kf_lib_data_ingest/etl/configuration/target_api_config.py
```diff
@@ -9,8 +9,8 @@ class Foo:
         class_name = 'foo'
         target_id_concept = CONCEPT.FOO.TARGET_SERVICE_ID
 
-        @staticmethod
-        def transform_records_list(records_list):
+        @classmethod
```
liberaliscomputing, line 12: Do we have use cases where these methods need to be factories for the class?

kf_lib_data_ingest/etl/load/load_v2.py
```diff
@@ -0,0 +1,79 @@
...
+        tic = record.get(entity_class.target_id_concept)
+        if tic and (tic != constants.COMMON.NOT_REPORTED):
+            return tic
+
+        # check the cache
+        try:
+            key_components = self._do_target_get_key(entity_class, record)
+            tic = self._get_target_id_from_key(
+                entity_class.class_name, str(key_components)
+            )
```
liberaliscomputing, line 35: After this block, just

```python
return tic
```

kf_lib_data_ingest/etl/load/load_v2.py
```diff
@@ -0,0 +1,79 @@
...
+        except Exception:
+            return None
+
+        if tic:
+            return tic
```
liberaliscomputing, line 40: Delete this block and do as above?

kf_lib_data_ingest/etl/load/load_v2.py
```diff
@@ -0,0 +1,79 @@
...
+        :rtype: str
+        """
+        # check if target ID is given
+        tic = record.get(entity_class.target_id_concept)
+        if tic and (tic != constants.COMMON.NOT_REPORTED):
+            return tic
+
+        # check the cache
+        try:
+            key_components = self._do_target_get_key(entity_class, record)
```
liberaliscomputing, line 32: In V1, a unique key was a concatenation of fields by `"|"` and V2 stores a stringified dictionary, right?

kf_lib_data_ingest/etl/configuration/target_api_config.py
fiendish: Not factories, but some of the methods do refer to the class itself (not all, though) and I wanted to switch to using `cls` instead of the specific declaration because I felt like it added some visual clarity.

Example:
Visually differentiating between
`"kf_id": get_target_id_from_record(cls, record),`     <- self class
and 
`"investigator_id": get_target_id_from_record(Investigator, record),`   <- other class
on https://github.com/kids-first/kf-lib-data-ingest/pull/513/files/a4cb66e73a110d852f66b5cfa63b3276438e48bb#diff-443b5b7d88394f11f7ccf7a8fc95383d265626dcaedb88e62d2043f7c0384248R89-R90

kf_lib_data_ingest/app/cli.py
fiendish: > When this flag is set, does the ingest library remove an existing UID cache

Yes.

> and send HTTP requests to the server to 1) GET the first batch's UIDs

Yes, but only once per entity because it also repopulates the cache with the response results.

> so it wouldn't create new resources

I assume you mean for ones that overlap with the first batch. In theory, yes.

> and 2) POST the second batch resources?

Yes.

---

> When unset, does the ingest library only send POST requests for the second batch resources?

It will send post requests for anything in the current batch because like always the non-key information could have changed.

kf_lib_data_ingest/etl/load/load_v2.py
fiendish: Stringified result from get_key_components. Whether it's a dictionary depends on the plugin.

kf_lib_data_ingest/etl/load/load_v2.py
fiendish: Then it would never check the server.

kf_lib_data_ingest/etl/load/load_v2.py
fiendish: Then it would never check the server.

kf_lib_data_ingest/app/cli.py
liberaliscomputing: Clearly explained! I am also down for the idea of wiping out an existing cache for every run. This way, we can conisder the data in FHIR as a source of truth, not heavily relying a cache (somethimes deprecated).

kf_lib_data_ingest/etl/load/load_v2.py
liberaliscomputing: Ah, the above logic is checking whether a key-components uid already exists.

liberaliscomputing:
@fiendish, ping me after rebasing. I will approve then.

liberaliscomputing (approved):
✅ Looks good to get merged in!

Issue #516, opened 2020-10-22
Imports in Extract Config example in the docs should import Split()


chris-s-friedman:
The example extract config in the docs uses `Split()` but does not import Split(). The docs do [mention](https://kids-first.github.io/kf-lib-data-ingest/tutorial/extract/index.html?highlight=split#the-split-object) where the function comes from in (`common.pandas_utils.Split`) but the example never imports the function.

PR #517, opened 2020-10-22
📝 import Split() in docs extract config

chris-s-friedman:
Closes https://github.com/kids-first/kf-lib-data-ingest/issues/516


znatty22 (approved):
👍

PR #518, opened 2020-10-22
📝 Consolidate example extract config copies

fiendish:


chris-s-friedman:
Now that i'm thinking about it - does `from kf_lib_data_ingest.etl.extract.operations import *` automatically import `Split` because of https://github.com/kids-first/kf-lib-data-ingest/blob/master/kf_lib_data_ingest/etl/extract/operations.py#L17-L21 ?

fiendish:
Yes but I don't mind

fiendish:
honestly, my linter even complains about importing * instead of explicit entries, but I'm ok with it

chris-s-friedman (approved):
👍

Issue #512, opened 2020-10-08
Update ingest docs to incorporate new validation 


znatty22:
- Remove old validation stuff having to do with accounting
- Update quick-start and how-to tutorial etc with new validation

PR #523, opened 2020-11-09
📝 Update documentation for recent changes

fiendish:
Installation instructions should be clearer.
New validation flags/output.
Refactored target service plugin API.
Ingest packages should be sub-components of studies.


liberaliscomputing:
Re  `docs/source/_static/images/ingest-pipeline.png`,  `kids_first.py` inputs to Transform and Load (in terms of the arrow direction), but isn't the config actually moving one way toward Load?

docs/source/design/load.rst
```diff
@@ -27,56 +27,51 @@ Each builder class implements methods for the following behaviors:
...
-   assigned to the same entities again. (Target services often don't do this
-   themselves.)
-#. **Build a message payload that will be sent to the server to populate a new
-   (or updated) entity of its type**. The unique keystrings generated in the
-   previous steps are passed to this method as well so that they can be
-   included as part of the payload if desired.
+#. **Compose the entity's uniquely identifying primary elements**. This is used
+   for querying identifiers assigned by the target service so that they can be
+   repeatably assigned to the same entities again. (Target services often don't
+   do this equivalent entity replacement themselves.)
```
liberaliscomputing, line 33: As far as I understand, the sequential operation of `get_key_components` and `query_target_ids` would also create new target service IDs when entities don't exist. The statement here sounds a bit like this operation is only for replicating what's in the target service.

docs/source/design/load.rst
```diff
@@ -27,56 +27,51 @@ Each builder class implements methods for the following behaviors:
...
+be: ``{"study_id": "SD_12345678", "external_id": "P1"}`` because every
+participant external ID is guaranteed to be unique only within a given study.
+Any future corrections to the participants age or sex should be associated with
+the same entry.
+
+Let's say that, after submission to the target service, the returned identifier
+for the participant entity is something like: ``"PT_00001111"``.
+
+The ingest library will internally create an association relating this unique
+key with its target identifier like this:
```
liberaliscomputing, line 63: Isn't what creates the internal association a target API builder (ingest plugin) with the ingest library invoking this builder? Since we imagine ingest plugins against other target services are developed separately from the ingest library, why don't we clarify that a little bit here?

docs/source/index.rst
```diff
@@ -77,10 +64,17 @@ Users of the ingest system will likely fall into 2 categories:
 
 2. **Ingest Package Developer**
 
-   - Runs ingest packages
```
liberaliscomputing, line 80: Doesn't this statement still need to be here? In most cases (and not in the near future), an ingest package developer would always be a ingest operator as well. Then, "Will never create a new study ingest package" seems conflicting the roles of an ingest package developer.

docs/source/index.rst
```diff
@@ -133,78 +128,27 @@ Extract stage output might look like this:
...
-                }
-            }
-        ]
-    }
-
-
+The transform stage applies the user-defined transform function which merges
+the individual tables created by the Extract Stage into one or more "records"
+tables that each contain all of the descriptive information for an instance of
+a real world entity or event on each row.
```
liberaliscomputing, line 134: This new statement looks enough sufficient for me. I wonder if it's enough informative for newcomers though.

docs/source/quickstart.rst
```diff
@@ -105,8 +117,8 @@ which is by default located at ``<ingest package dir>/logs/ingest.log``.
 View single stage output
 ------------------------
 
-Every ingest stage has the option to write its output to a directory that
-follows this path pattern: ``<ingest package dir>/output/<name of stage>``.
+Ingest stages write their output to directories that follow this path pattern:
```
liberaliscomputing, line 120: We may want to clarify these outputs are in a tabular and/or JSON format?

docs/source/design/load.rst
fiendish: The target service itself creates new IDs. Querying just determines if one already exists.

docs/source/quickstart.rst
```diff
@@ -183,32 +194,34 @@ out:
...
-If the ingest package passes validation with the test command and you're happy
-with the output, you're ready to ingest the data into your target service using
-the ingest command::
+The Warehouse Server
+--------------------
 
-    $ kidsfirst ingest my_study
+Data from ingesting is meant to be shared, not stored locally on your machine!
+It's easy to configure the ingest system to connect to a remote warehousing
+database which will centrally store intermediate stage data.
```
liberaliscomputing, line 224: Why don't we mention how to export a warehouse URL so that it can be grabbed by the ingest library?

docs/source/tutorial/extract/extract_example.inc
```diff
@@ -53,9 +53,9 @@ The following extract configuration file accomplishes all of those needs
 It goes in the extract_configs directory like this::
 
     my_study/
-    ├── data
-    └── extract_configs
-        └── family_and_phenotype.py
+    └── my_ingest_package/
```
liberaliscomputing, line 56: `ingest_1` for consistency across the doc?

docs/source/tutorial/package_setup.rst
```diff
@@ -4,40 +4,25 @@
...
-        └── test_custom_counts.py
-
-Try running your ingest package like this to make sure it works:
+Create an outer directory for ``my_study`` and inside it a new ingest package
+named ``my_ingest_package``:
 
 .. code-block:: text
 
-    $ kidsfirst test my_study
+    $ kidsfirst new --dest_dir=my_study/my_ingest_package
```
liberaliscomputing, line 12: `ingest_1` for consistency?

docs/source/tutorial/package_setup.rst
```diff
@@ -4,40 +4,25 @@
...
-    ======================================== 1 passed in 0.12s ============================================================
-    2020-02-18 17:40:36,039 - DataIngestPipeline - Thread: MainThread - INFO - ✅ User defined data validation tests passed
-    2020-02-18 17:40:36,039 - DataIngestPipeline - Thread: MainThread - INFO - END data ingestion
-    2020-02-18 17:40:36,039 - kf_lib_data_ingest.app.cli - Thread: MainThread - INFO - ✅ Ingest pipline passed validation!
+This will generate a minimal study directory for you that looks like this::
 
-You'll learn more about the ``test`` and ``ingest`` commands later in
-:ref:`Tutorial-Package-Testing`.
+    my_study/
+    └── my_ingest_package/
```
liberaliscomputing, line 17: `ingest_1` for consistency?

docs/source/tutorial/package_testing.rst
```diff
@@ -11,19 +11,20 @@ At this point, your ingest package should have the source data file and extract
...
-    │   └── clinical.tsv
-    ├── extract_configs
-    │   ├── extract_config.py
-    │   └── family_and_phenotype.py
-    ├── tests
-    │   ├── conftest.py
-    │   └── test_custom_counts.py
-    ├── ingest_package_config.py
-    └── transform_module.py
+    └── my_ingest_package/
```
liberaliscomputing, line 17: `ingest_1` for consistency?

docs/source/tutorial/package_testing.rst
```diff
@@ -32,7 +33,7 @@ Let's see what happens if you run:
 
 .. code-block:: text
 
-    $ kidsfirst test my_study
+    $ kidsfirst test my_study/my_ingest_package
```
liberaliscomputing, line 36: `ingest_1` for consistency?

docs/source/tutorial/package_testing.rst
```diff
@@ -64,13 +65,14 @@ transform module merge function OR your transform module merge function is
...
-Try uncommenting the commented lines in ``my_study/transform_module.py``
+Try uncommenting the commented lines in
+``my_study/my_ingest_package/transform_module.py``
 
 Now test your package again:
 
 .. code-block:: text
 
-    $ kidsfirst test my_study
+    $ kidsfirst test my_study/my_ingest_package
```
liberaliscomputing, line 75: `ingest_1` for consistency?

docs/source/tutorial/target_service_plugins/how_to_make.rst
```diff
@@ -18,28 +18,37 @@ Here are two complete examples that work with the same input data for different
...
+       of the plugin API they conform to by defining a variable
+       ``LOADER_VERSION = <version number>`` in the plugin body.
+     | Example: ``LOADER_VERSION = 2``
 
-#. **The submit function**
+#. **The list of entity builder classes**
 
-   - This negotiates sending the payload to the target service.
+   - These classes are responsible for converting lists of records into
+     entity payloads, querying the target server for existing entities that
```
liberaliscomputing, line 48: `entity payloads, querying the target server for existing entities that`

>>

`entity payloads, getting the unique key components from the entities, querying the target server for existing` entities that`?

znatty22 (changes requested):
I'm seeing fairly minor changes for the addition of data validation and old references to expected counts, concept discovery, etc. I think the old stuff has to be removed and some more detail should be added to the data validation section. The docs should point the user to where the validation reports get generated, maybe show a screenshot of the markdown report, and explain one or two validation rules.

docs/source/design/load.rst
```diff
@@ -27,56 +27,51 @@ Each builder class implements methods for the following behaviors:
...
-always point to the phenotype observation entity in the target service
-identified by ``"PH_00001111"``.
+   :header: STUDY.TARGET_SERVICE_ID, PARTICIPANT.ID, PARTICIPANT.AGE, PARTICIPANT.SEX
+
+   SD_12345678,  P1, 17, Male
+
+The uniquely identifying components for the participant in this record would
+be: ``{"study_id": "SD_12345678", "external_id": "P1"}`` because every
+participant external ID is guaranteed to be unique only within a given study.
+Any future corrections to the participants age or sex should be associated with
```
gsantia, line 56: `Any future corrections to the participant's age or sex should be associated with`

docs/source/install.rst
```diff
@@ -2,15 +2,20 @@
...
 
-Install the current release of ``kf-lib-data-ingest`` with ``pip``::
+    $ pip install psycopg2-binary
 
-    $ pip install -e git+https://github.com/kids-first/kf-lib-data-ingest.git#egg=kf-lib-data-ingest
+If a precompiled binary for psycopg2 doesn't exist for your platform, or
+if you prefer not to use one, installation with pip may not be straightforward
+everywhere (see e.g.
+https://web.archive.org/web/20190908200859/https://stackoverflow.com/questions/26288042/error-installing-psycopg2-library-not-found-for-lssl),
+so we leave it as an exercise for the reader.
```
gsantia, line 14: 😂

docs/source/design/load.rst
fiendish: The api builder returns the important parts, but the library itself contains the cache relation.

docs/source/index.rst
fiendish: I think a person can have multiple roles.

fiendish: > Then, "Will never create a new study ingest package" seems conflicting the roles of an ingest package developer.

Yeah, I'll remove it.

> Doesn't this statement still need to be here? In most cases (and not in the near future), an ingest package developer would always be a ingest operator as well. 

I think removing the above makes it safe to say that people can have multiple roles.

docs/source/quickstart.rst
fiendish: I want to avoid indicating format because then it needs to change if the format changes.

docs/source/quickstart.rst
fiendish: I don't have any insight into this. Getting our warehouse db URL is very convoluted.

liberaliscomputing: Oh, I simply meant `export KF_WAREHOUSE_DB_URL=<dialect>://<user>:<password>@<host>:<port>/<database>`.

znatty22 (changes requested):
Can you please move the validation report builder changes into a different PR?

fiendish:
I will. I'm not done here and github doesn't let you move back to draft mode.

fiendish:
Oh, I was wrong. You _can_ move back to draft mode. The link for it is just tiny.

docs/source/design/load.rst
```diff
@@ -27,56 +27,51 @@ Each builder class implements methods for the following behaviors:
...
+| participants                                                     |
++--------------------------------------------------+---------------+
+| Unique Components                                | Target ID     |
++==================================================+===============+
+| {"study_id": "SD_12345678", "external_id": "P1"} |  PT_00001111  |
++--------------------------------------------------+---------------+
+
+Then when the ingest library sees a record in the source data with the same
+unique key components, it can always associate the resulting entity in the
+target service with the same target ID. That way we can update characteristics
```
znatty22, line 75:
```suggestion
target service with the same target service ID. That way we can update characteristics
```

docs/source/tutorial/testing.rst
```diff
@@ -0,0 +1,161 @@
...
+determine whether the output of each stage matches a predefined set of
+relationship and value expectations.
+
+The analysis has a basic understanding of the internal hierarchical
+relationships captured by the data model of our target service (participants
+are grouped into families, biospecimens come from individual participants, etc)
+as well as constraints on the values that each data field can have (e.g. age
+can't be negative).
+
+It will try to evaluate
```
znatty22, line 64: Left over sentence that shouldn't be here?

znatty22 (changes requested):

znatty22 (approved):
💯 📝

fiendish:
closes #512

PR #524, opened 2020-11-10
👽 Add auth configs for new study creator subdomains

### LINKED ISSUE #525: New study creator subdomains break FileRetriever default configs because auth isn't set
Auth configs are matched by subdomain, so now that the study creator reports kf-api-study-creator instead of kf-study-creator for the download url, downloading is broken.

fiendish:
closes #525


znatty22 (approved):
👍 ...I wouldn't emojify it with a bug though since its not _our_ doing that caused the breakage 😛

PR #527, opened 2020-11-11
🐛 Don't show blank cells as invalid values

### LINKED ISSUE #526: Blank cells are treated as invalid values
As long as data validation works directly on the dataframes and not on the graph (I forget why it does this. Faster maybe?), it shouldn't treat blank cells as invalid values. It should just ignore them as unpopulated.

fiendish:
closes #526


znatty22 (approved):
👍

PR #529, opened 2020-11-11
🐛 Fix bad call order blocking BS and SE loads

### LINKED ISSUE #528: Bad call order in target service plugin blocks loading BS and SE with missing replaceable fields
BS aliquot ID should fill in a missing sample group ID.
SE library should fill in missing ID and vice versa.

Using `r.get(a, b)` instead of `r.get(a) or b` is breaking the replacement if b doesn't exist.

fiendish:
closes #528


znatty22 (approved):
💯

PR #531, opened 2020-11-11
🐛 Fake-cache target IDs in dry run mode 

### LINKED ISSUE #530: Dry run mode doesn't dry-load relationships
It looks like the cause is not caching keys during dry runs, so then when the loader tries to build the relationship entity the primary components (nominally kfids) are missing.

fiendish:
Needed for dry-loading relationship entities, because otherwise their
primary components can't be built.

closes #530


znatty22 (approved):
🥇

PR #532, opened 2020-11-18
⬆️ Update dependencies for Python 3.9

fiendish:
The big one is xlrd which is broken in 3.9 for a terrible reason.


znatty22 (changes requested):
We should probably update the circleci image to use the python 3.9 image (if there is one?) too right?

znatty22 (dismissed):
🆒 ✅

znatty22 (approved):
🍪 👍

PR #537, opened 2020-12-01
🔊 Add html validation report

fiendish:
Generated from the markdown with some github-ish css. This is for #512 because the other report formats require special tools to be useful to humans.


znatty22 (commented):
Looks good except for the issue with the content enclosed in back ticks not being rendered properly. Should be good to go after that's fixed in the `markdownescape` branch

znatty22 (approved):
💯 Thanks for implementing this and the markdown escaping!

PR #549, opened 2020-12-14
✨ Convert family relations to P1+P2+rel

fiendish:
We have two formats for family relationship. We often get PID+relationship_to_proband, but we need P1+P2+relationship.


znatty22 (commented):
✅

liberaliscomputing (approved):

PR #552, opened 2020-12-18
✨ Add option to query from another source in Load

fiendish:
Using kidsfirst `test`/`ingest --dry-run` (or loading into a local private dataservice), you can add `--clear_cache --query_url=https://kf-api-dataservice.kidsfirstdrc.org` to query against the actual prod dataservice to find matching production kfids to see in the logs exactly what will get added vs updated.

@chris-s-friedman asked for this and I thought it was a good idea


liberaliscomputing (approved):
Looks good!

PR #554, opened 2021-01-14
🐛 Ensure integers fit into allocated bits

fiendish:
This chooses to null ones that are too big instead of assigning the wrong value.

### DIFF
--- a/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
+++ b/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
@@ -8,8 +8,9 @@ First Dataservice)
 See etl.configuration.target_api_config docstring for more details on the
 requirements for format and content.
 """
+import logging
+
 from d3b_utils.requests_retry import Session
-from pandas import DataFrame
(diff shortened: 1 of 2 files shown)


kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -874,7 +874,15 @@ def coerce_types(host, entity_class, body):
                 # Don't include it so that the server applies default behavior.
                 continue
         else:
-            ret[k] = json_type_casts.get(properties[k]["type"], lambda x: x)(v)
+            v = json_type_casts.get(properties[k]["type"], lambda x: x)(v)
+            if properties[k]["type"] == "integer":
+                try:
+                    positive_bits = int(properties[k]["format"][-2:]) - 1
+                    if v > 2 ** positive_bits:
+                        v = None
```
znatty22, line 882: Should we log some sort of warning when a value becomes null?

fiendish: added

znatty22 (approved):
👍

Issue #540, opened 2020-12-03
Possible issue with sqlite3worker on WSL2


torstees:
I think there may be some timing issue with the sqlite3worker on WSL. I have had weird issues with ids that simply refuse to make it to the database even though the insert statement was called (and thus the next time the ingestion was run, those without IDs were treated as new and end up with duplicate entries in FHIR). 

Would it be possible to make that database be based on plain sqlite3 based on an option in the __init__?

fiendish:
I think this _should_ be possible. Minimally it will require changing when and where entries get written to the cache db.

blackdenc:
@torstees Can you share your `/etc/wsl.conf` file? I don't think we officially support WSL/WSL2, but there are some settings you may need to change there that might help with the issue.

torstees:
Sure, I'll paste it below, since it's not particularly long:

blackdenc:
So this is just a guess, but I think the issue you're having might be related to `root = /mnt`. The WSL2 I/O performance is only on the Linux filesystem, it doesn't extend to the NTFS drives under `/mnt` so it's possible that the cache lookup is missing due to that. 

https://docs.microsoft.com/en-us/windows/wsl/compare-versions#performance-across-os-file-systems

torstees:
Well, I need files available to both windows and WSL with my current workflow when it comes to data ingestion and the entire system used for it lives in NTFS and will continue to be that way as long as I use Windows as my primary OS. However, I do have a work around if it doesn't seem appropriate to officially provide a solution.

blackdenc:
You can still access the NTFS filesystem under `/mnt` without defining it as the root, but the application and dependencies should be installed under `/` in the Linux FS to avoid performance issues. I'm not sure if that'll work for your use-case, and most of the time cross-OS executables works fine, but in cases like this I'd still recommend running databases and application code inside the Linux VM.

fiendish:
I do plan to make changes to remove the sqlite queueing thread, because fewer dependencies is usually better, but it's not clear yet how funneling requests into a queue because sqlite can't handle parallel writes (because it has no queueing process of its own) causes entries to be missed only on WSL.

torstees:
[...]

PR #555, opened 2021-01-19
➖ Use a mutex instead of sqlite3worker

fiendish:
closes https://github.com/kids-first/kf-lib-data-ingest/issues/540


fiendish:
Not sure I should have merged those other PRs into this one, but oh well, haha

znatty22 (approved):
✅ Seems like this would work lol

PR #556, opened 2021-01-20
♻️ Put V2 kfid queries inside async threads

fiendish:


kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -840,17 +842,19 @@ def _GET(host, api_path, body):
...
 def coerce_types(host, entity_class, body):
-    if not swagger_cache:
-        swagger = get_open_api_v2_schema(host)
-        defs = swagger["definitions"]
-        for c in all_targets:
-            n = c.class_name
-            uccn = upper_camel_case(n)
-            if uccn in defs:
-                swagger_cache[n] = defs[uccn]
+    with swag:
```
znatty22, line 849: Why is this needed?

fiendish: So that only the first thread bothers to populate the cache instead of requesting swagger.json a dozen times in a row.

znatty22: ooo i see

znatty22 (approved):
✅

PR #558, opened 2021-01-20
🥅 Error if read DF is empty

### LINKED ISSUE #557: Cryptic error message if do_after_read returns an empty DF
Calling row_map on an empty df after do_after_read removed all rows gave the cryptic error "Wrong number of items passed 7, placement implies 1" from deep inside pandas and a key error on the output column. The problem is not obvious from the messaging.

fiendish:
closes #557


znatty22 (approved):
✅

PR #563, opened 2021-01-26
🔥 Remove test files from template project

fiendish:
Remove the test stuff from the template project until it becomes useful again. Right now the files are just clutter.


znatty22 (approved):
I'm fine with this as long as no one is actively using the test stuff

fiendish:
Even if they are, the test stuff doesn't get run anymore.

PR #564, opened 2021-01-26
🐛 Validate V1 API configurations

### LINKED ISSUE #560: target api config validator doesn't check version
@torstees reports that the target api config validator no longer accepts version 1 configurations. That is true and a bug.

fiendish:
closes #560

### DIFF
--- a/kf_lib_data_ingest/etl/configuration/target_api_config.py
+++ b/kf_lib_data_ingest/etl/configuration/target_api_config.py
@@ -5,6 +5,8 @@ Your target API configuration module must contain a list named `all_targets`
 which contains, in the order that you want them to be loaded into the target
 service, target entity builder classes (not instances) of the form:
 
+[if LOADER_VERSION == 2]
+
     class Foo:
         class_name = 'foo'
         target_id_concept = CONCEPT.FOO.TARGET_SERVICE_ID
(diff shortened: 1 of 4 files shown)


liberaliscomputing (approved):
✅ Looks good to get merged in!

kf_lib_data_ingest/etl/configuration/target_api_config.py
```diff
@@ -5,6 +5,8 @@
 which contains, in the order that you want them to be loaded into the target
```
znatty22, line 5: I'll just put this comment up here since it applies to the loader v1 and v2 docstrings. Its hard to know which methods of the class are required vs not. The only one I know is not required is the transform_records_list bc the docstring says so

znatty22 (changes requested):

kf_lib_data_ingest/etl/configuration/target_api_config.py
fiendish: Hmm. Proposed fix?

znatty22: How about adding something at the top of each method's docstring? `OPTIONAL` or `REQUIRED`. I know we talked about this but I forget the explanation. Why not use abstract classes to mandate what should be implemented?

fiendish: > Why not use abstract classes to mandate what should be implemented?

As far as I know, abc methods don't enforce call signatures.

znatty22 (approved):
👍

PR #565, opened 2021-01-26
✨ Add consent_type and dbgap_consent_code to BS

chris-s-friedman:
the kids first api has consent type and dbgap consent code as attributes of biospecimen. although these values are assigned elsewhere (see https://github.com/kids-first/kf-update-dbgap-consent) for non-dbgap studies or for studies where dbgap has not released the data, these values can only be set in standalone scripts. adding these two items to the biospecimen concept allows these to be added during ingest.

### DIFF
--- a/kf_lib_data_ingest/common/concept_schema.py
+++ b/kf_lib_data_ingest/common/concept_schema.py
@@ -143,6 +143,8 @@ class CONCEPT:
         CONCENTRATION_MG_PER_ML = None
         VOLUME_UL = None
         SAMPLE_PROCUREMENT = None
+        DBGAP_STYLE_CONSENT_CODE = None
+        CONSENT_SHORT_NAME = None
 
     class GENOMIC_FILE(PropertyMixin, FileMixin):
         HARMONIZED = None
--- a/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
+++ b/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
@@ -425,6 +425,10 @@ class Biospecimen:
             "method_of_sample_procurement": record.get(
                 CONCEPT.BIOSPECIMEN.SAMPLE_PROCUREMENT
             ),
+            "dbgap_consent_code": record.get(
+                CONCEPT.BIOSPECIMEN.DBGAP_STYLE_CONSENT_CODE
+            ),
+            "consent_type": record.get(CONCEPT.BIOSPECIMEN.CONSENT_SHORT_NAME),
         }
         return {
             **cls.get_key_components(record, get_target_id_from_record),


fiendish:
> for studies where dbgap has not released the data

If the data is not on dbgap then where is dbgap data coming from?

Please figure out what "duo_ids" are and report back on how they interact with either of these.
https://github.com/kids-first/kf-api-dataservice/blob/98f69e747f72a1b780b748f8d713a57c01c41111/dataservice/api/biospecimen/models.py#L115-L118

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -138,6 +138,8 @@ class EVENT_AGE(QuantityMixin):
         CONCENTRATION_MG_PER_ML = None
         VOLUME_UL = None
         SAMPLE_PROCUREMENT = None
+        DBGAP_CONSENT_CODE = None
```
fiendish, line 141: Is there anything dbgap-specific about the format of these codes?

chris-s-friedman: related to your other comment

> > for non-dbgap studies or for studies where dbgap has not released the data
> 
> Where is dbgap data coming from if not from dbgap? If it's coming from somewhere else then it sounds like it isn't dbgap anything.

you're correct. this could be any type of consent code. perhaps it was a mistake to name it `DBGAP_CONSENT_CODE` - referencing the target api. So a better name would be?

```suggestion
        CONSENT_CODE = None
```

fiendish: I don't know. I think that depends on why the dataservice calls its field "dbgap".

fiendish: New question. Is there only one code per specimen?

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -395,6 +395,10 @@ def build_entity(cls, record, get_target_id_from_record):
             "method_of_sample_procurement": record.get(
                 CONCEPT.BIOSPECIMEN.SAMPLE_PROCUREMENT
             ),
+            "dbgap_consent_code": record.get(
+                CONCEPT.BIOSPECIMEN.DBGAP_CONSENT_CODE
```
liberaliscomputing, line 399: It still uses `DBGAP_CONSENT_CODE`, not `CONSENT_CODE`

kf_lib_data_ingest/common/concept_schema.py
chris-s-friedman: yes - only one code per specimen

liberaliscomputing: As far as I remember, the assignment of specimen-level consent code is dbGaP-specific (I am not sure about whether it can have multiple values).

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
chris-s-friedman: correct - I haven't made the change yet. wasn't clear if we wanted that change

kf_lib_data_ingest/common/concept_schema.py
liberaliscomputing: I am actually down for `CONSENT_CODE` since it is inclusive of other scenarios. Why don't we double-check with Allison and Bailey?

fiendish:
@chris-s-friedman What's the status of this? I want to make a release.

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
chris-s-friedman:
```suggestion
                CONCEPT.BIOSPECIMEN.CONSENT_CODE
```

chris-s-friedman:
@fiendish i've updated so that the library itself no longer referes to dbgap consent code - but rather `CONCEPT.BIOSPECIMEN.CONSENT_CODE` is connected to dbgap consent code in the target api plugin for kidsfirst

fiendish (changes requested):
Open issues:

1. I don't like using a generic identifier for something specific to a particular service. If this is for _any_ type of consent code, then it seems inappropriate to put it into a dataservice field that isn't for _any_ type of consent code without interpretation. If you call it just "consent code" then you need another thing that identifies what kind, and at that point you're back to just calling it what it is in the first place, a dbgap-style consent code.

2. I still don't see why codes from dbgap aren't being applied using the existing tool specifically designed for doing that which follows a long list of procedural rules that are not being enforced here.

3. CONSENT_TYPE is not descriptive. CONSENT_SHORT_NAME would be better.

4. What's going on with duo_ids?

chris-s-friedman:
@fiendish now that some time has passed and i've had some time to think about it, this PR is a response to weird things that we do to put non-KF data in the KF dataservice. This PR is really about handling how `we` - both d3b and the kids first portal ` handle genomic files and biospecimens from the kids first dataservice. 

So:

>     1. I don't like using a generic identifier for something specific to a particular service. If this is for _any_ type of consent code, then it seems inappropriate to put it into a dataservice field that isn't for _any_ type of consent code without interpretation. If you call it just "consent code" then you need another thing that identifies what kind, and at that point you're back to just calling it what it is in the first place, a dbgap-style consent code.

^ This is *the* thing - the dataservice has non-dbgap style consent codes in the dbgap-consent-code field so that other tools - like the portal's variant workbench which queries *both* acl and dbgap consent code- can work correctly for non dbgap studies. For example CBTN biospecimens have entries in dbgap consent code so those specimens show up in the variant workbench. 

>     2. I still don't see why codes from dbgap aren't being applied using the existing tool specifically designed for doing that which follows a long list of procedural rules that are not being enforced here. The dbgap consent updater does a lot more than just set codes, and it does so because the desired response to different consents is more complex than just populating a field on biospecimen.

see above - this is for codes that are not from dbgap but are going in the dataservice dbgap consent code field. 

>     3. CONSENT_TYPE is not descriptive. CONSENT_SHORT_NAME would be better.

will change

>     4. What's going on with duo_ids?

not a clue - other than that field not being used by a single biospecimen in the dataservice.

fiendish (approved):
Yay!

PR #567, opened 2021-01-26
🐛 Don't nullify coerced integers

fiendish:
The dataservice could be lying about the sizes of fields like it does for the genomic file size field.
So instead of nulling the field, print a guiding message in case the next submission fails.


znatty22 (approved):
🟢

PR #568, opened 2021-01-27
♻️ Don't silently cross-assign SE.SL/SE.ID

fiendish:
It's probably a bad idea to silently wedge values into other slots even if in theory they should be 1:1. 
This will force the operator to assign CONCEPT.SEQUENCING.ID.

### DIFF
--- a/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
+++ b/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
@@ -505,17 +505,11 @@ class SequencingExperiment:
     @classmethod
     def get_key_components(cls, record, get_target_id_from_record):
         return {
+            "study_id": not_none(record[CONCEPT.STUDY.TARGET_SERVICE_ID]),
             "sequencing_center_id": not_none(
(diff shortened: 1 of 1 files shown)


kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -505,17 +505,11 @@ class SequencingExperiment:
...
             ),
-            "external_id": (
-                record.get(CONCEPT.SEQUENCING.ID)
-                or not_none(record[CONCEPT.SEQUENCING.LIBRARY_NAME])
-            ),
-            "library_name": (
-                record.get(CONCEPT.SEQUENCING.LIBRARY_NAME)
-                or not_none(record[CONCEPT.SEQUENCING.ID])
-            ),
+            "external_id": not_none(record[CONCEPT.SEQUENCING.ID]),
```
liberaliscomputing, line 512: I totally agree that we shouldn't rely on other attributes even if we know that they are a 1-on-1 match.

znatty22 (approved):
Yea I agree ✅

liberaliscomputing (approved):
✅ Let's merge!

PR #570, opened 2021-02-03
🔊 🐛 More verbose loader logging + loader fixes

### LINKED ISSUE #545: Need more insight into skipped loads
Trying to load biospecimen_diagnosis without also loading biospecimen, diagnosis, participant doesn't appear to work but doesn't say why. We need to indicate when things are being skipped and the reasons why.

### LINKED ISSUE #569: More graceful handling of GF missing required attributes
There are attributes/columns in the Kids First Target API config that are required and when they're missing in the dfs, the resulting exception obscures this. For example: 

https://github.com/kids-first/kf-lib-data-ingest/blob/f84e27264f2ae35c8ff5adf826fa3e8cc37b3880/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py#L445

Here file size is a required attribute on genomic file. If it is missing the int conversion on NoneType will result in a TypeError. It would be better to error and report that a required field is missing or set it to None and let dataservice tell us a required field is missing.

znatty22:
Tested locally

Closes #545 
Closes #569

### DIFF
--- a/kf_lib_data_ingest/etl/load/load_v1.py
+++ b/kf_lib_data_ingest/etl/load/load_v1.py
@@ -72,7 +72,7 @@ class LoadStage(IngestStage):
             when using resume_from, because that needs the cache to be effective.
         :type clear_cache: bool, optional
         """
-        super().__init__(cache_dir)
+        super().__init__(cache_dir or os.getcwd())
         self.target_api_config = TargetAPIConfig(target_api_config_path)
         self._validate_entities(
             entities_to_load,
@@ -87,7 +87,7 @@ class LoadStage(IngestStage):
 
         target = urlparse(target_url).netloc or urlparse(target_url).path
         self.uid_cache_filepath = os.path.join(
-            cache_dir or os.getcwd(),
+            self.stage_cache_dir,
             #  Every target gets its own cache because they don't share UIDs
             "_".join(multisplit(target, [":", "/"]))
             #  Every study gets its own cache to compartmentalize internal IDs
@@ -266,15 +266,25 @@ class LoadStage(IngestStage):
         """
         try:
             key_components = self._do_target_get_key(entity_class, record)
-            unique_key = str(key_components)
         except Exception:
             # no new key, no new entity
+            key_components = None
+
+        if not key_components:
+            self.logger.debug(
+                f"Skip {entity_class.class_name}. Missing key components. "
+                f"Failed to construct unique key from record:"
+                f"\n{pformat(record)}"
+            )
             return
 
-        if (not key_components) or (
-            unique_key in self.seen_entities[entity_class.class_name]
-        ):
+        unique_key = str(key_components)
+        if unique_key in self.seen_entities[entity_class.class_name]:
             # no new key, no new entity
+            self.logger.debug(
+                f"Skip {entity_class.class_name}. Duplicate record found in "
+                f"data:\n{record}"
+            )
             return
 
         self.seen_entities[entity_class.class_name].add(unique_key)
@@ -379,7 +389,8 @@ class LoadStage(IngestStage):
             for entity_class in self.target_api_config.all_targets:
                 if entity_class.class_name not in self.entities_to_load:
                     self.logger.info(
-                        f"Skipping load of {entity_class.class_name}"
+                        f"Skipping load of {entity_class.class_name}. Not "
+                        "included in ingest package config."
                     )
                     continue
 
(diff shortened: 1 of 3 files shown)


kf_lib_data_ingest/etl/load/load_v2.py
```diff
@@ -50,12 +50,15 @@ def _get_target_id_from_record(self, entity_class, record):
...
+
+        err = False
         try:
             tic_list = entity_class.query_target_ids(
                 self.query_url or self.target_url, key_components
             )
             if tic_list:
                 if len(tic_list) > 1:
+                    err = True
                     raise Exception(
```
znatty22, line 62: This was never getting raised because of the return in the try/except

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -668,9 +678,16 @@ def query_target_ids(cls, host, key_components):
 
     @classmethod
     def build_entity(cls, record, get_target_id_from_record):
+        def external_id(record):
```
znatty22, line 681: Could be in another PR, sorry 😬 , but I snuck these external_id additions into each of the many-to-many association tables. I like seeing the external_id field filled in for these in Data Service because it gives me some idea of which two things are related from a source data perspective. Pretty much for debugging purposes or if I need to do a delete or bulk delete. For example seeing `external_id:  Schiffman-SS01-file_abc12345.bam` gives me an idea of which specimen and file are related in this BG record.

fiendish: I kinda think these should be using a string version of otherclass.get_key_components(record, get_target_id_from_record) no? biospecimen ID alone doesn't uniquely identify a specimen

fiendish (changes requested):

kf_lib_data_ingest/etl/load/load_v1.py
```diff
@@ -87,11 +87,13 @@ def __init__(
...
         self.uid_cache_filepath = os.path.join(
-            cache_dir or os.getcwd(),
             #  Every target gets its own cache because they don't share UIDs
             "_".join(multisplit(target, [":", "/"]))
             #  Every study gets its own cache to compartmentalize internal IDs
-            + "_" + study_id + "_" + DEFAULT_ID_CACHE_FILENAME,
+            + "_"
+            + study_id
+            + "_"
+            + DEFAULT_ID_CACHE_FILENAME,
```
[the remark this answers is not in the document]
znatty22: yea sorry my auto formatter did that 🤦🏾‍♀️

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
znatty22: Mm hmm yea I think you're right

znatty22: Oh no, nevermind. That results in values like this for `external_id: PT_TP0FPEXC-BS0-BS0-SD_ME0WME0W-GF0`. I want the values from the source data

kf_lib_data_ingest/etl/load/load_v1.py
```diff
@@ -87,11 +87,13 @@ def __init__(
 
         target = urlparse(target_url).netloc or urlparse(target_url).path
         self.uid_cache_filepath = os.path.join(
-            cache_dir or os.getcwd(),
```
fiendish, line 90: why remove this?

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
fiendish: Hmm. What about putting participant_id in build_entity instead of key components on biospecimen and replace it with study_id. Then you can just remove the study_id field here before concatenating values.

kf_lib_data_ingest/etl/load/load_v1.py
znatty22: Wellll, bc when you instantiate loader with cache_dir=None, it throws an exception bc the base Stage class uses it

znatty22: Oh also, its not removed. I just moved setting cache_dir up in the constructor

fiendish: but it looks like this doesn't use the cache dir anymore

znatty22: Ohhh yea sorry I see what you're saying now. I accidentally removed that little important line altogether.

fiendish (approved):
👑

Issue #573, opened 2021-02-08
"Build and publish" github workflow never runs


fiendish:
It looks like it blocks on a job that no longer exists in the file:
https://github.com/kids-first/kf-lib-data-ingest/blob/6f2e6b1c0f7d448e938d582e4df5d6c6fe321672/.github/workflows/publish.yml#L9

PR #574, opened 2021-02-09
💚 Remove requirement for job in testing workflow

blackdenc:
Fixing issue https://github.com/kids-first/kf-lib-data-ingest/issues/573


fiendish (approved):
Thanks!

PR #576, opened 2021-02-14
✨ Autogenerate family relations

### LINKED ISSUE #323: Family pedigree should be as dense as possible
Complex families sometimes come in with labels like "maternal grandmother", "twin sister", etc. Siblings should be linked to the parents. Labeled grandparents should be linked to the appropriate parents. The links are never explicit in the data and must be filled in.

I have scattered code for this. It's pretty wild, some of it probably overlaps with other parts, and it will definitely need to be adapted.

reference:
https://github.com/kids-first/kf-study-imports/blob/5fa02e766b3a631311aa1edc40544a246bdb20d8/Avi/transform.py#L186
https://github.com/kids-first/kf-study-imports/blob/5fa02e766b3a631311aa1edc40544a246bdb20d8/Avi/load.py#L604
https://github.com/kids-first/kf-study-imports/blob/5fa02e766b3a631311aa1edc40544a246bdb20d8/Avi/keys_and_codes.py#L163-L285

fiendish:
This tries to automatically adapt from [FAMILY, PID, RELATIONSHIP_TO_PROBAND] and [PID, MOTHER_ID, FATHER_ID] into [P1, RELATIONSHIP_OF_1_TO_2, P2].

It also tries to add some unstated but inferrable relationships like that the proband's sister is also the proband's mother's daughter.

It also checks for gender/relationship mismatch errors.

closes #323 

This may not necessarily stay in the api plugin, but I don't think that matters since the code is all in its own module.


liberaliscomputing (approved):
✅ We went through the implementation details together, adding minor fixes including the comments about "twin" only indicating "identical." I approve this PR without further concerns!

PR #580, opened 2021-02-24
✨ Enable loading un-preregistered Studies

### LINKED ISSUE #577: Google's FHIR requires GUIDs for ID (and overwrites anything you provide for ID at create time)
For study, the ingestion library forces the use of a cleaned version of the study's name (as best as I can tell) to be used for the ID field for actual study resources and also references back to them. This is done by way of adding the transformed study's name the the record under the target_service_id.

When testing out my ingest package using google's healthcare fhir service, those IDs were stripped out and GUID were used instead, resulting in the all attempts to reference the study to fail (since we were assuming that ResearchStudy/study_id would be an actual FHIR resource. 

Removing the following lines from load_v1.py (413-414) resolved the issue for me. I would be happy to issue a PR, but I honestly don't know what sort of implications this may have on other parts of the library. 

`
                # guarantee existence of the study ID column
                for r in transformed_records:
                    r[CONCEPT.STUDY.TARGET_SERVICE_ID] = self.study_id
`

I believe there are other ways to reference an object, however, using those IDs is the cleanest. The current approach works fine for the hapi FHIR server, but it seems google is a bit picker.

fiendish:
Not everyone pre-registers their studies, so we shouldn't rely on Study target service identifiers being available. 

closes #577

### DIFF
--- a/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
+++ b/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
@@ -98,19 +98,27 @@ class Study:
 
     @classmethod
     def get_key_components(cls, record, get_target_id_from_record):
-        return {
-            "data_access_authority": not_none(record[CONCEPT.STUDY.AUTHORITY]),
-            "external_id": not_none(record[CONCEPT.STUDY.ID]),
-        }
+        kfid = record.get(cls.target_id_concept) or record.get(
+            CONCEPT.PROJECT.ID
+        )
+        au = record.get(CONCEPT.STUDY.AUTHORITY)
+        id = record.get(CONCEPT.STUDY.ID)
+        assert (au and id) or kfid
+        return {"kf_id": kfid, "data_access_authority": au, "external_id": id}
 
     @classmethod
     def query_target_ids(cls, host, key_components):
-        return list(yield_kfids(host, cls.api_path, drop_none(key_components)))
+        kfid = key_components.pop("kf_id", None)
+        if kfid:
+            return [kfid]
+        else:
+            return list(
+                yield_kfids(host, cls.api_path, drop_none(key_components))
+            )
 
     @classmethod
     def build_entity(cls, record, get_target_id_from_record):
         secondary_components = {
-            "kf_id": get_target_id_from_record(cls, record),
             "investigator_id": get_target_id_from_record(Investigator, record),
             "name": record.get(CONCEPT.STUDY.NAME),
             "short_name": record.get(CONCEPT.STUDY.SHORT_NAME),
@@ -139,7 +147,7 @@ class Family:
     @classmethod
     def get_key_components(cls, record, get_target_id_from_record):
         return {
-            "study_id": not_none(record[CONCEPT.STUDY.TARGET_SERVICE_ID]),
+            "study_id": get_target_id_from_record(Study, record),
             "external_id": not_none(record[CONCEPT.FAMILY.ID]),
         }
 
@@ -172,7 +180,7 @@ class Participant:
     @classmethod
     def get_key_components(cls, record, get_target_id_from_record):
         return {
-            "study_id": not_none(record[CONCEPT.STUDY.TARGET_SERVICE_ID]),
+            "study_id": get_target_id_from_record(Study, record),
             "external_id": not_none(record[CONCEPT.PARTICIPANT.ID]),
         }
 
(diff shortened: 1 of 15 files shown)


fiendish:
I have confirmation that this appears to work.

znatty22 (changes requested):
Some lingering references to study

```
➜  kf-lib-data-ingest git:(independentresearch) ✗ rg study kf_lib_data_ingest

kf_lib_data_ingest/validation/data_validator.py
46:        logger.info("Validating study data")
110:        """Construct a graph that represents the study data. Nodes are cells in
373:        rules have been broken by the study data.

kf_lib_data_ingest/templates/my_ingest_package/ingest_package_config.py
26:# TODO - Replace this with your own unique identifier for the study. This will

kf_lib_data_ingest/factory/generate.py
33:    # Don't overwrite existing study dir

kf_lib_data_ingest/etl/ingest_pipeline.py
11:    init_study_db,
12:    persist_df_to_study_db,
251:                study_id = self.data_ingest_config.project
252:                init_study_db(self.warehouse_db_url, study_id, self.ingest_name)
254:                    f"Initialized db in warehouse for {study_id} {self.ingest_name}."
288:                        schema = persist_df_to_study_db(
297:                            f"Loaded {schema}.{df_name} in {study_id} warehouse."

kf_lib_data_ingest/etl/extract/extract.py
70:            <study config dir path>/output_cache/<ingest stage class name>.xxx

kf_lib_data_ingest/common/warehousing.py
11:def db_study_url(db_maintenance_url, study_id):
12:    return urlparse(db_maintenance_url)._replace(path=f"/{study_id}").geturl()
15:def init_study_db(db_maintenance_url, study_id, ingest_package_name):
17:    Create and/or reset a warehouse database for the given study.
21:    :param study_id: Unique identifier for the study
22:    :type study_id: str
24:    # create the study database if needed
32:        eng.execute(f'CREATE DATABASE "{study_id}"')
38:    # link to the study database
40:        db_study_url(db_maintenance_url, study_id),
52:def persist_df_to_study_db(
54:    study_id,
64:    :param study_id: Unique identifier for the study
65:    :type study_id: str
73:    # link to the study database
75:        db_study_url(db_maintenance_url, study_id),
```

znatty22 (dismissed):
🐣

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -98,19 +98,27 @@ class Study:
...
-        return {
-            "data_access_authority": not_none(record[CONCEPT.STUDY.AUTHORITY]),
-            "external_id": not_none(record[CONCEPT.STUDY.ID]),
-        }
+        kfid = record.get(cls.target_id_concept) or record.get(
+            CONCEPT.PROJECT.ID
+        )
+        au = record.get(CONCEPT.STUDY.AUTHORITY)
+        id = record.get(CONCEPT.STUDY.ID)
+        assert (au and id) or kfid
```
liberaliscomputing, line 106: Referring to the dataservice SD spec (https://kf-api-dataservice.kidsfirstdrc.org/#tag/Study%2Fpaths%2F~1studies%2Fpost), `data_access_authority` is not required?

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -98,19 +98,27 @@ class Study:
 
     @classmethod
     def get_key_components(cls, record, get_target_id_from_record):
-        return {
-            "data_access_authority": not_none(record[CONCEPT.STUDY.AUTHORITY]),
-            "external_id": not_none(record[CONCEPT.STUDY.ID]),
-        }
+        kfid = record.get(cls.target_id_concept) or record.get(
```
liberaliscomputing, line 101: Since an unregistered project ID is not necessarily a KF ID, we may rename this variable?

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
fiendish: I'm just worried that study external id may not be unique :(

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
fiendish: it's being assigned to the kf_id field. should it not?

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
liberaliscomputing: ugh, hmm. That is a good point under the assumption that, within the same access authority, a study id should be unique?

fiendish: that's my hope! haha

liberaliscomputing (approved):

PR #582, opened 2021-03-04
✨ Allow some extract operations to be optional

### LINKED ISSUE #579: Outside suggestion: extract operations conditional on applicability (e.g. in_col existence)
Does anyone have any thoughts on whether this is a fundamentally good thing or bad thing for the ingest library to include?

---

There has been a request from the data tracker team for extract operations to be skipped without error (or at least not raise an uncaught exception) if conditions make them inapplicable so that the same set of operations can be reused for files with different subsets of known _optional_ columns.

I definitely don't want extract to continue after other errors, so if this is implemented I think it would require applicability checking in each operation.

Ignoring df and row maps that operate on entire frames and constant map that operates without a reference, the remaining global operations are:

**value**
potential condition: in_col exists

**column**
potential condition: in_col exists

**keep**
potential condition: in_col exists

**melt**
potential condition: _all_ map_for_vars columns exist

---

**The obvious benefit** is the ability to use static predefined configurations for files with different column subsets without having to dynamically generate the operations lists. 

**The obvious problem** is that skipping over invalid entries will make it easier to miss and thus harder to debug invalid configurations. 

Possible mitigation strategies for this concern:

* Add another argument to those operations to designate them as explicitly "optional" and only allow skipping them if set to True.
* Logging `warning` level messages when skipping an operation for invalidity.
* Including the skipped operations and the reasons for skipping them in the final run output.

**The obvious alternative** is for the data tracker to compose operations list dynamically and not depend on the proposed behavior.

fiendish:
closes #579

### DIFF
--- a/kf_lib_data_ingest/etl/extract/operations.py
+++ b/kf_lib_data_ingest/etl/extract/operations.py
@@ -22,6 +22,11 @@ from kf_lib_data_ingest.common.pandas_utils import (  # noqa: F401
 from kf_lib_data_ingest.common.type_safety import assert_safe_type
 
 
+class SkipOptional:
+    def __init__(self, needed_columns):
+        self.needed_columns = needed_columns
+
+
 def df_map(m):
     """
     Wraps the DataFrame mapping operation in a function that takes only a
@@ -46,7 +51,7 @@ def df_map(m):
     return df_map_func
 
 
-def value_map(m, in_col, out_col):
+def value_map(m, in_col, out_col, optional=False):
     """
     Wraps the value mapping operation in a function that takes a DataFrame as
     input and returns a single column DataFrame (plus index).
@@ -81,6 +86,8 @@ def value_map(m, in_col, out_col):
 
     def value_map_func(df):
         new_df = DataFrame()
+        if optional and (in_col not in df):
+            return SkipOptional([in_col])
         if callable(m):
             new_df[out_col] = get_col(df, in_col).apply(m)
         elif isinstance(m, str):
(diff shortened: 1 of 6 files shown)


kf_lib_data_ingest/etl/extract/operations.py
```diff
@@ -81,6 +86,8 @@ def value_map(m, in_col, out_col):
 
     def value_map_func(df):
         new_df = DataFrame()
+        if optional and (in_col not in df):
```
znatty22, line 89: Shouldn't there be an error if the operation is NOT optional (required) and the `in_col` is not found?

kf_lib_data_ingest/etl/extract/operations.py
```diff
@@ -22,6 +22,11 @@
 from kf_lib_data_ingest.common.type_safety import assert_safe_type
 
 
+class SkipOptional(BaseException):
```
znatty22, line 25: Can this be renamed to something more intuitive or something that sounds more like an error? If I just see `raise SkipOptional` I don't think I could even guess what that exception is for

znatty22 (dismissed):

kf_lib_data_ingest/etl/extract/operations.py
fiendish: there will be

kf_lib_data_ingest/etl/extract/operations.py
fiendish: So I'm not even sure that I should use an exception here. I could instead return a <thing> and check if isinstance(ret, <thing>) instead of raising and catching.

fiendish:
Ok I replaced the raise with a return to make it more clear that this is not an error condition.

dankolbman (approved):
# 🗼
lgtm

PR #587, opened 2021-03-10
🐛 Don't error on relation external_id without key fields

fiendish:
If the record is missing requisite key fields for the two related
sub-entities in a relation entity (BSGF, etc), just skip populating the
external_id field instead of raising an exception.

### DIFF
--- a/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
+++ b/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
@@ -44,16 +44,19 @@ def not_none(val):
 
 
 def external_id(cls_list, record, get_target_id_from_record):
-    out = []
-    for cls in cls_list:
-        cls_key_dict = cls.get_key_components(
-            record, get_target_id_from_record
-        ).copy()
-        for key in ["study_id", "sequencing_center_id"]:
-            cls_key_dict.pop(key, None)
-        out.extend([str(v) for v in cls_key_dict.values()])
-
-    return DELIMITER.join(out) or None
+    try:
+        out = []
+        for cls in cls_list:
+            cls_key_dict = cls.get_key_components(
+                record, get_target_id_from_record
+            ).copy()
+            for key in ["study_id", "sequencing_center_id"]:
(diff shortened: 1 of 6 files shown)


chris-s-friedman (approved):

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -44,16 +44,19 @@ def not_none(val):
...
-        out.extend([str(v) for v in cls_key_dict.values()])
-
-    return DELIMITER.join(out) or None
+    try:
+        out = []
+        for cls in cls_list:
+            cls_key_dict = cls.get_key_components(
+                record, get_target_id_from_record
+            ).copy()
+            for key in ["study_id", "sequencing_center_id"]:
```
znatty22, line 53: Why did we do this again?

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -44,16 +44,19 @@ def not_none(val):
...
+                cls_key_dict.pop(key, None)
+            out.extend([str(v) for v in cls_key_dict.values()])
+    except Exception:
```
znatty22, line 56: I've been bit by these catch all exceptions recently and its made debugging really hard. Is it possible to catch a more specific exception?

fiendish: It will probably be safe to catch just KeyError. It wasn't clear to me that any exception should create an error here.

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
fiendish: I think you just didn't want those fields used from SE for SEGF .

znatty22: Ohh right 👍

znatty22 (approved):
Cool, looks good!

PR #592, opened 2021-03-22
✨ Standalone Extractor

### LINKED ISSUE #593: Standalone Extractor Utility
In preparation for automating validation in Study Creator we've realized we can't use the ingest lib ExtractStage as-is to execution extraction. We need to be able to "apply" an extract config to a given DataFrame to produce the extracted df. 

The reason we want to operate on a DataFrame is because Study Creator has its own functionality for fetching a file (whether on local file system or on S3) and converting to a DataFrame. This functionality also takes care of any authentication needed. 

The ExtractStage assumes the input data is the form of remotely stored files (local, S3, http, etc) and most of the logic outside of the actual extraction logic has to do with setting up output directories and fetching the remote files which may or may not require authentication. 

Ideally, it would be nice to have a standalone extract utility that can do this:
- Extractor().extract(df, extract_config_or_filepath) --> Study Creator use case
- Extractor().extract(df_or_filepath, extract_config_or_filepath) --> Maybe some other user wants this?

See this WIP code:

https://github.com/kids-first/kf-api-study-creator/blob/3b5b27b6efdddb7dc83b141ef5ebe9477e6a0d2d/creator/ingest_runs/validation/scratch.py#L76

that kinda shows how we want to use this magical extraction utility.

znatty22:
Closes #593

### DIFF
--- a/kf_lib_data_ingest/etl/extract/utils.py
+++ b/kf_lib_data_ingest/etl/extract/utils.py
@@ -1,0 +1,294 @@
+from collections import defaultdict
+import logging
+from functools import reduce
+from math import gcd
+
+import pandas
+
+from kf_lib_data_ingest.common.concept_schema import concept_set
+from kf_lib_data_ingest.common.misc import clean_up_df
+from kf_lib_data_ingest.common.type_safety import (
+    assert_safe_type,
+    is_function,
+)
+from kf_lib_data_ingest.common.pandas_utils import split_df_rows_on_splits
+from kf_lib_data_ingest.etl.configuration.base_config import (
+    ConfigValidationError,
+)
+from kf_lib_data_ingest.etl.configuration.extract_config import ExtractConfig
+from kf_lib_data_ingest.etl.extract import operations as extract_operations
+
+
+def lcm(number_list):
+    """
+    Returns the least common multiple from a list of numbers.
+    """
+    return reduce(lambda x, y: x * y // gcd(x, y), number_list)
+
+
+def ordinal(n):
+    """
+    Convert a positive integer into its ordinal representation
+    """
+    suffix = ["th", "st", "nd", "rd", "th"][min(n % 10, 4)]
+    if 11 <= (n % 100) <= 13:
+        suffix = "th"
+    return str(n) + suffix
+
+
+class Extractor(object):
+    """
+    Encapsulates the functionality to clean and standardize a source DataFrame
+    according to the define operations in the extract configuration file for
+    the DataFrame.
+
+    Used internally by the ExtractStage or can be used as a standalone utility.
+    """
+
+    def __init__(self, logger=None):
+        self.logger = logger or logging.getLogger(type(self).__name__)
+        self.messages = []
+
+    def _log_operation(self, op, nth):
+        """Log execution of an extract operation.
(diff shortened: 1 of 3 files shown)


tests/test_extract_stage.py
```diff
@@ -103,6 +104,24 @@ def test_extracts():
             pandas.testing.assert_frame_equal(A, B)
 
 
+def test_extractor_standalone():
```
znatty22, line 107: Will implement rest of tests once we agree on implementation

fiendish (approved):

fiendish:
Looks good, but please lint first.

znatty22:
@fiendish Lol its not done yet. Still have to add a couple more tests and yes, have to do some dusting

kf_lib_data_ingest/etl/extract/utils.py
```diff
@@ -0,0 +1,292 @@
...
+        df_out, file_skips = self._chain_operations(
+            df_in, extract_config.operations
+        )
+
+        # record skipped operations
+        if file_skips:
+            self.messages.append(
+                f"In file {extract_config.config_file_relpath}:"
+            )
+            self.messages.extend(f"\t{s}" for s in file_skips)
```
fiendish, line 280: Now if you use the same Extract instance multiple times the internal messages list will get longer and longer. That appears to not be intentional since the ExtractStage uses the same instance over and over. Changing this to be a local messages list here instead of self.messages and returning it to the caller at the end of this method would align more closely with the previous operation.

fiendish, line 280: It looks like now if you use the same Extract instance multiple times the internal messages list will get longer and longer. That appears to not be intentional since the ExtractStage uses the same instance over and over and extends the outer message list each time. Changing this to be a local messages list here instead of self.messages and returning it to the caller at the end of this method and extending the outer message list with the returned list would align more closely with the previous operation.

fiendish (changes requested):

kf_lib_data_ingest/etl/extract/utils.py
znatty22: I can't return messages since the extract function needs to return the extracted df. Instead I just reset `self.messages = []` at the top of extract and then extract stage can access it the way it does now.

fiendish: You can return more than one thing.

Like...previously ExtractStage._run() would call _chain_operations many times, once for each extract configuration. And now ExtractStage._run() will call Extractor.extract many times, once for each extract configuration, in pretty much exactly the same way.

So if Extractor.extract is just a thin wrapper around _chain_operations, then it could return the same way that _chain_operations does.

But anyway this works too. I just think it looks a little funny to have an instance variable inside of Extractor for something that doesn't preserve state across invocations. But then I think that making Extractor a class at all instead of just a module function is kinda weird too. Not weird enough to not approve the change though.

fiendish (approved):
approved with commentary

znatty22:
@fiendish Yea I know its a little weird. But now that extractor is something that will be used outside ingest library, I thought returning more than just the extracted df would be weirder for the user.

PR #597, opened 2021-04-16
✨ New functionality to support Data Tracker templates

znatty22:
Needed to close https://github.com/kids-first/kf-api-study-creator/pull/571

- Adds ontology code extraction methods
- Adds new concepts to support the next generation of Kids First data templates here:
https://docs.google.com/spreadsheets/d/1ugcw1Rh3e7vXnc7OWlR4J7bafiBjGnfd4-rEThI-BNI/edit#gid=408414488


kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -102,20 +102,32 @@ class DIAGNOSIS(PropertyMixin):
         CATEGORY = None
         UBERON_TUMOR_LOCATION_ID = None
         EVENT_AGE_DAYS = None
+        VERIFICATION = None
+
+        class ABATEMENT_EVENT_AGE(QuantityMixin):
```
fiendish, line 107: I think we should have the corresponding _DAYS field with any new age quantities to use with https://github.com/kids-first/kf-lib-data-ingest/blob/92e8b9224a32a82d63e655af25a356c144b4da69/kf_lib_data_ingest/common/misc.py#L18-L31

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -133,9 +145,18 @@ class BIOSPECIMEN(PropertyMixin):
         COMPOSITION = None
         EVENT_AGE_DAYS = None
 
+        class COLLECTION_EVENT_AGE(QuantityMixin):
```
fiendish, line 148: is this different than the existing "EVENT_AGE"?

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -102,20 +102,32 @@ class DIAGNOSIS(PropertyMixin):
...
+
+        HPO_ID = None
         MONDO_ID = None
         NCIT_ID = None
         ICD_ID = None
 
     class PHENOTYPE(PropertyMixin):
         NAME = None
         HPO_ID = None
+        MONDO_ID = None
```
fiendish, line 126: mondo is for diseases not phenotypes :|

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -102,20 +102,32 @@ class DIAGNOSIS(PropertyMixin):
...
 
         class EVENT_AGE(QuantityMixin):
             pass
 
+        class ANATOMY_SITE:
+            NAME = None
+            NCIT_ID = None
+            UBERON_ID = None
+
+        HPO_ID = None
```
fiendish, line 118: hpo is for phenotypes and not diseases

znatty22:
cc @allisonheath 

@fiendish Yea I agree with your comments, but I've been following whats in the file type guidelines here: https://www.notion.so/d3b/File-Types-and-Guidelines-fc5bd4390fa54de5a70b550d73779de9. Maybe we need to review these again and make sure they're still accurate?

kf_lib_data_ingest/common/concept_schema.py
znatty22: Why though? I thought the other instances of this were for backward compatibility?

kf_lib_data_ingest/common/concept_schema.py
znatty22: Hmm..idk..probably not

kf_lib_data_ingest/common/concept_schema.py
liberaliscomputing: MONDO IS for diseases, not phenotypic features.

kf_lib_data_ingest/common/concept_schema.py
liberaliscomputing: Yes, HPO is for phenotypic features.

kf_lib_data_ingest/common/constants.py
```diff
@@ -14,6 +14,8 @@ class COMMON:
     NOT_AVAILABLE = "Not Available"
     NOT_REPORTED = "Not Reported"
     NO_MATCH = "No Match"
+    # New: Needed to support Data Tracker templates
+    NOT_PROVIDED = "Not Able to Provide"
```
fiendish, line 18: I think I would favor
```suggestion
    NOT_ABLE_TO_PROVIDE = "Not Able to Provide"
```
since not provided and not able to provide don't mean the same thing to me

kf_lib_data_ingest/validation/values.py
```diff
@@ -11,6 +11,7 @@
 MAX_AGE_DAYS = 32872
 NULL_VALUES = {
     NA,
+    constants.COMMON.NOT_PROVIDED,
```
fiendish, line 14: here too

fiendish (changes requested):

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -102,10 +108,19 @@ class DIAGNOSIS(PropertyMixin):
...
+            pass
 
         class EVENT_AGE(QuantityMixin):
             pass
 
+        class ANATOMY_SITE:
+            NAME = None
+            NCIT_ID = None
+            UBERON_ID = None
+
```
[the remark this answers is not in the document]
znatty22: I think we don't need `ANATOMY_SITE.NCIT_ID`. I'll remove that.

In general, I've taken the approach of keeping old fields and adding new fields in an effort to ease the transition from our current Dataservice model to a future model (whether a modification of Dataservice or a new FHIR service). Then in the Data Tracker extract configs, I tried to map a template column to both an "old"/current concept (because those are used in the current target service plugin) and the new concept (so that the extracted files will have the new fields that can be used by a future target service plugin).  Not sure if that's the best approach but that's why I have `ANATOMY_SITE` concepts and the other location concepts

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -136,6 +176,12 @@ class BIOSPECIMEN(PropertyMixin):
         class EVENT_AGE(QuantityMixin):
             pass
 
+        class VOLUME(QuantityMixin):
+            pass
+
+        class CONCENTRATION(QuantityMixin):
+            pass
+
```
fiendish, line 184: I think we determined that volume/concentration aren't the exact right concepts for specimens. Good luck finding the relevant comment stream on notion, though. I don't know where it is or when it was written.

kf_lib_data_ingest/common/constants.py
```diff
@@ -14,6 +14,8 @@ class COMMON:
     NOT_AVAILABLE = "Not Available"
     NOT_REPORTED = "Not Reported"
     NO_MATCH = "No Match"
+    # New: Needed to support Data Tracker templates
```
fiendish, line 17:
```suggestion
```

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -116,6 +130,32 @@ class PHENOTYPE(PropertyMixin):
...
+        class EVENT_AGE(QuantityMixin):
+            pass
+
+        class ANATOMY_SITE:
+            NAME = None
+            UBERON_ID = None
+
+    # NOTE: Not yet used by any target service plugin, but needed for
+    # the Data Tracker templates + extract configs used to maximize data
+    # captured from investigators
```
fiendish, line 148:
```suggestion
```

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -71,6 +71,12 @@ class PARTICIPANT(PropertyMixin):
         PROBAND_ID = None
         RELATIONSHIP_TO_PROBAND = None
         GENDER = None
+
+        # NOTE: Not yet used in target service, but needed in order to capture
+        # gender identity in Data Tracker templates
```
fiendish, line 76:
```suggestion
```

I don't want comments that have to be deleted

kf_lib_data_ingest/common/concept_schema.py
```diff
@@ -116,6 +130,32 @@ class PHENOTYPE(PropertyMixin):
...
+
+        class ABATEMENT_EVENT_AGE(QuantityMixin):
+            pass
+
+        class EVENT_AGE(QuantityMixin):
+            pass
+
+        class ANATOMY_SITE:
+            NAME = None
+            UBERON_ID = None
```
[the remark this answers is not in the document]
znatty22: This is because we're using the condition resource to represent both phenotypes and diagnoses. The condition resource has anatomy site

fiendish (approved):

PR #618, opened 2021-06-07
🐛 Assign IDs to new entities in dry_run mode that can be re-used during the run

### LINKED ISSUE #613: relationship entities are not being created during testing
when running kidsfirst test, relationship entities aren't being created because there are no target ids being created

fiendish:
closes #613


gsantia (approved):
yess

znatty22 (approved):

PR #622, opened 2021-06-15
🚸 Warehousing is no longer default behavior

### LINKED ISSUE #612: Should warehousing default to off?
I want to start a discussion about this: I think warehousing should default to off. 

Unless I have a special case where I *need* to warehouse the output of my ingest run, i always run `kidsfirst test` and `kidsfirst ingest` with the `--no_warehouse` flag. 

I'm curious if any of the other ingest package developers are warehousing the outputs of their ingests and how they feel about defaulting warehousing to be off.

fiendish:
closes #612

### DIFF
--- a/kf_lib_data_ingest/app/cli.py
+++ b/kf_lib_data_ingest/app/cli.py
@@ -197,10 +197,10 @@ def cli():
     "the target service. Overrides the resume_from setting.",
 )
 @click.option(
-    "--no_warehouse",
+    "--warehouse",
     default=False,
     is_flag=True,
-    help="Skip sending data to the warehouse db.",
+    help="Send data to the warehouse db.",
 )
 def ingest(
     ingest_package_path,
@@ -211,7 +211,7 @@ def ingest(
     use_async,
     dry_run,
     resume_from,
-    no_warehouse,
+    warehouse,
     no_validate,
     validation_mode,
     clear_cache,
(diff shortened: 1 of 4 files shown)


kf_lib_data_ingest/app/cli.py
```diff
@@ -197,10 +197,10 @@ def cli():
     "the target service. Overrides the resume_from setting.",
 )
 @click.option(
-    "--no_warehouse",
+    "--warehouse",
     default=False,
```
znatty22, line 201: This means the `warehouse` kwarg is set to `False`  by default right? I can't remember click

gsantia (approved):
┌༼ຈل͜ຈ༽┘

kf_lib_data_ingest/app/cli.py
fiendish: It _should_ mean that :D

Christina-J-Diaz:
✨ ✅

Christina-J-Diaz (approved):

kf_lib_data_ingest/app/cli.py
fiendish: I think a boolean toggle must always default to false, so it probably doesn't even need to be there. Like if it defaulted to true, how would you set it false right?

znatty22 (approved):

Issue #629, opened 2021-08-18
Unexpected list instead of dict during LoadStage


gsantia:
While running an ingest, @Christina-J-Diaz encountered the following error:
```
2021-08-18 16:31:03,910 - GuidedTransformStage - Thread: MainThread - INFO - ❌ TransformStage failed validation!
2021-08-18 16:31:03,910 - GuidedTransformStage - Thread: MainThread - INFO - END GuidedTransformStage. Time elapsed: Sec: 0.9110019207000732 Min: 0.0 Hours: 0.0
2021-08-18 16:31:03,920 - LoadStage - Thread: MainThread - INFO - Loader Version: 2
2021-08-18 16:31:03,920 - LoadStage - Thread: MainThread - INFO - BEGIN LoadStage
2021-08-18 16:31:03,921 - LoadStage - Thread: MainThread - INFO - Skipping load of investigator. Not included in ingest package config.
2021-08-18 16:31:03,921 - LoadStage - Thread: MainThread - INFO - Skipping load of study. Not included in ingest package config.
2021-08-18 16:31:03,921 - LoadStage - Thread: MainThread - INFO - Skipping load of family. Not included in ingest package config.
2021-08-18 16:31:03,921 - LoadStage - Thread: MainThread - INFO - Skipping load of participant. Not included in ingest package config.
2021-08-18 16:31:03,921 - LoadStage - Thread: MainThread - INFO - Skipping load of family_relationship. Not included in ingest package config.
2021-08-18 16:31:03,921 - LoadStage - Thread: MainThread - INFO - Skipping load of diagnosis. Not included in ingest package config.
2021-08-18 16:31:03,921 - LoadStage - Thread: MainThread - INFO - Skipping load of phenotype. Not included in ingest package config.
2021-08-18 16:31:03,921 - LoadStage - Thread: MainThread - INFO - Skipping load of outcome. Not included in ingest package config.
2021-08-18 16:31:03,921 - LoadStage - Thread: MainThread - INFO - Skipping load of biospecimen. Not included in ingest package config.
2021-08-18 16:31:03,921 - LoadStage - Thread: MainThread - INFO - Skipping load of genomic_file. Not included in ingest package config.
2021-08-18 16:31:03,921 - LoadStage - Thread: MainThread - INFO - Skipping load of read_group. Not included in ingest package config.
2021-08-18 16:31:03,922 - LoadStage - Thread: MainThread - INFO - Begin loading sequencing_experiment
[...]

PR #631, opened 2021-08-20
🐛 Raise in KF target when response is bad with good status

gsantia:
Partially addresses #629.

Raise a `RequestException` in the Kids First target API plugin when the response has a status code of 200 or 201 but is not in the form that is expected.

### DIFF
--- a/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
+++ b/kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
@@ -986,8 +986,13 @@ def submit(host, entity_class, body):
         resp = _POST(host, api_path, body)
 
     if resp.status_code in {200, 201}:
-        result = resp.json()["results"]
-        return result["kf_id"]
+        try:
+            return resp.json()["results"]["kf_id"]
+        except Exception as e:
+            raise RequestException(
+                f"Got status 200/201 from /{api_path} but response was not "
+                f"as expected:\nSent:\n{body}\nGot:\n{resp.text}"
+            ) from e
     elif (resp.status_code == 400) and (
         "already exists" in resp.json()["_status"]["message"]
     ):


kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -986,8 +987,14 @@ def submit(host, entity_class, body):
...
     if resp.status_code in {200, 201}:
-        result = resp.json()["results"]
-        return result["kf_id"]
+        result = resp.json()
+        if isinstance(result, dict) and "results" in result:
+            return result.get("kf_id")
+        else:
+            raise RequestException(
+                f"Got response code 200/201 but response is not as expected."
+                f"Response was: {pprint(result)}"
```
chris-s-friedman, line 996: Looks great! How does this message look in the logs? i'm just curious if it would look cleaner if this line was split into two lines? I honestly don't know if it even matters though.... 
```suggestion
                f"Response was:"
                pprint(result)
```

chris-s-friedman (commented):
looks good. I'm curious what that result looks like in the log? i'm wondering if

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
gsantia: how about
```
                f"Got response code 200/201 but response is not as expected."
                f"Response was: \n{pprint(result)}"
```

kf_lib_data_ingest/target_api_plugins/kids_first_dataservice.py
```diff
@@ -986,8 +987,14 @@ def submit(host, entity_class, body):
...
-        result = resp.json()["results"]
-        return result["kf_id"]
+        result = resp.json()
+        if isinstance(result, dict) and "results" in result:
+            return result["results"].get("kf_id")
+        else:
+            raise RequestException(
+                f"Got response code 200/201 but response is not as expected."
+                f"Response was:\n{pformat(result)}"
+            )
```
fiendish, line 997: If the response is coming back in an unexpected format, we don't know that it will parse as json. What about this?
```suggestion
        try:
            return resp.json()["results"]["kf_id"]
        except Exception as e:
            raise RequestException(
                f"Got status 200/201 from /{api_path} but response was not as expected:"
                f"\nSent:\n{body}\nGot:\n{resp.text}"
           ) from e
```

gsantia: much better

fiendish (approved):
🎯

PR #635, opened 2021-09-16
🐛 Fix VISIBLE/HIDDEN obversion to work with strings as expected

### LINKED ISSUE #634: flip_col doesn't flip False when it is a string
In an ingest i'm currently writing, I noticed that my extract config was creating a dataframe with values for both visible _and_ hidden for the concept i was applying a value of visible to: 

```python
constant_map(
        m=constants.COMMON.FALSE, out_col=CONCEPT.GENOMIC_FILE.VISIBLE
    ),
```

What struck me was that, although i set visible to `False`, both visible and hidden were set to false when i looked at the resulting table from this extract config. 

I think I found where the value of the inverse of hidden or visible is set, in kf_lib_data_ingest.etl.extract.utils.Extractor._obvert_visibility: https://github.com/kids-first/kf-lib-data-ingest/blob/ddc2f3e99ad40a56aa8af8141c96934d25fb7420/kf_lib_data_ingest/etl/extract/utils.py#L73-L92

I tested out the function, flip_col on it's own and think the issue is with how it handles False values that are strings. 

Take the following Data Frame: 

```python
df = pd.DataFrame(
    {
        "col_a":[False, "False", "0", 0, constants.COMMON.FALSE], 
        "col_b":[True, "True", "1", 1, constants.COMMON.TRUE]
    }
)
```

If I perform `flip_col` on `col_b`, everything works as expected where all the `True`'s - string or bool or int - are converted to `False`: 
```python
>>> flip_col(df['col_b'])
0    False
1    False
2    False
3    False
4    False
Name: col_b, dtype: object
```

Here's where my issue is: That doesn't work for when `False` is written as a string (as it is for `constants.COMMON.FALSE`) https://github.com/kids-first/kf-lib-data-ingest/blob/ddc2f3e99ad40a56aa8af8141c96934d25fb7420/kf_lib_data_ingest/common/constants.py#L21

`"False"`, `"0"`, and `constants.COMMON.FALSE` aren't being flipped to `True`
```python
>>> flip_col(df['col_a'])
0     True
1    False
2    False
3     True
4    False
Name: col_a, dtype: object
```

fiendish:
closes #634

### DIFF
--- a/kf_lib_data_ingest/etl/extract/utils.py
+++ b/kf_lib_data_ingest/etl/extract/utils.py
@@ -76,20 +76,60 @@ class Extractor(object):
         vice versa.
         """
 
+        def x_to_bool_or_none(x):
+            # Convert all obvious values to True/False/None and raise ValueError
+            # for any special cases, which are no longer allowed here.
+            if isinstance(x, bool):
+                return x
+            elif pandas.isna(x):
+                return None
+            elif isinstance(x, str):
+                lessx = x.strip().lower()
+                if not lessx:
+                    return None
+                elif lessx == "false":
+                    return False
+                elif lessx == "true":
+                    return True
+                else:
+                    raise ValueError(
+                        f'String must resolve to "false" or "true" or be empty. Got: "{x}"'
+                    )
+            else:
+                raise ValueError(
+                    "x_to_bool_or_none expects a bool, str, or pandas.isna(x)"
+                )
+
         def flip_col(in_col):
+            # Flip true/false for any non-null entries in this column
             out_col = in_col.copy()
-            out_col[out_col.notnull()] = (
-                ~in_col[in_col.notnull()].astype(bool)
-            ).astype(object)
+            out_col[out_col.notnull()] = ~(
+                in_col[in_col.notnull()].astype(bool)
+            )
             return out_col
 
         for concept in concept_set:
-            if concept.HIDDEN in df and concept.VISIBLE not in df:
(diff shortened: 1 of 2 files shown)


kf_lib_data_ingest/etl/extract/utils.py
```diff
@@ -76,20 +76,60 @@ def _obvert_visibility(self, df):
         vice versa.
         """
 
+        def x_to_bool_or_none(x):
+            # Convert all obvious values to True/False/None and raise ValueError
+            # for any special cases, which are no longer allowed here.
+            if isinstance(x, bool):
+                return x
+            elif pandas.isna(x):
```
znatty22, line 84:
```suggestion
            elif pandas.isnull(x):
```
Pandas says `isnull` also includes NaT (null datetime-like)? Other than that, idk what the difference is

znatty22 (approved):

kf_lib_data_ingest/etl/extract/utils.py
fiendish: I don't see any difference between
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.isna.html
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.isnull.html

fiendish: Also https://github.com/pandas-dev/pandas/blob/73c68257545b5f8530b7044f56647bd2db92e2ba/pandas/core/dtypes/missing.py#L141

PR #637, opened 2021-09-16
✨ Optionally exclude tests from validation report

### LINKED ISSUE #636: Optionally exclude tests in validation report
I would like a way to exclude relevant test sections from the validation report (e.g. only display relationship tests and count tests).

The Data Tracker is using the ingest library's data validator and reporting functionality for validating uploaded files and displaying results to the user. However, only the relationship test results are useful to display right now since the attribute tests rely on a set of hardcoded rules tailored to the Kids First program. In the future, we will be able to build the attribute test rules (e.g. Gender must be one of "Female, Male, Other") from the organization's template definitions and pass those into the validator. But this will require some refactoring in the data validator and some time to implement in Data Tracker.

It would be great if in the interim we can simply turn off the display of certain test sections in the report.

znatty22:
Closes #636

### DIFF
--- a/kf_lib_data_ingest/validation/reporting/markdown.py
+++ b/kf_lib_data_ingest/validation/reporting/markdown.py
@@ -53,7 +53,7 @@ class MarkdownReportBuilder(AbstractReportBuilder):
         """
         super().__init__(output_dir=output_dir, setup_logger=setup_logger)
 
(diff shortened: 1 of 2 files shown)


fiendish:
looks ok to me. still drafting?

znatty22:
@fiendish Oops forgot to undraft. Ready for approval!

fiendish (approved):
🐐

Issue #627, opened 2021-08-10
Allow extraction of data via passed in SQL query


gsantia:
Sometimes users will query the prod kf-postgres database for data using SQL and save the results into a format which the ingest library comprehends, then uses these files in an extract config. We could allow `.sql` files to be input to the extract stage and bypass this manual work.

Not entirely sure how credentials would work here, and if what steps may be needed to sanitize SQL statements before running them on production data.

chris-s-friedman:
so i'm thinking about a few things here, to do this, 

1. user passes source_data_url the connection url for whatever db they want to query, I.E. `postgresql://{USER_NAME}:{PASSWORD}@{PG_HOST}:{PG_PORT}/{PG_DBNAME}`
2. add functionality and changes to common.fileretriever to handle database protocols 
    A. add lines for different db protocols in common.file_retriever._getters
    B. add a getter for database connections. this would be something like _trying_ to connect to the database and if it fails, error out, but return `None`.
3. in common.io add a function read_sql_df that takes in the source_data_url and some sql and returns the results of pandas.read_sql(). 
4. make common.io.read_df know when to use read_sql_df

fiendish:
Why are we ingesting details that are already in the database?

chris-s-friedman:
This would be for pulling from any db-could be d3b warehouse. What's at top of my mind is pulling from file_metadata schéma in KFpostgres

fiendish:
> What's at top of my mind is pulling from file_metadata schéma in KFpostgres

Why?

chris-s-friedman:
Not having to run s3 scrapes manually since scrapes are now getting into the file_metadata.aws_scrapes table. Also assigning hashes from the file_metadata.hashes table

fiendish:
So you're putting s3 file metadata into its own table in the same database as the dataservice but not creating GFs in the dataservice so that they would get registered in indexd?

chris-s-friedman:
Correct. Not everything in s3 gets registered. That table allows comparing files between buckets and between accounts - e.g. allowing data ops to see if every file in a bucket is replicated in a DR bucket in another account- and the hashes of those files match. 

It allows seeing what files in s3 have not yet been registered. As well.

fiendish:
Anyway, as for the general problem...

[...]

PR #642, opened 2021-12-07
Sql query tutorial

chris-s-friedman:
add a tutorial on one method for sourcing data from a postgresql database and moved the tutorial on sourcing from study creator to the same directory as sourcing data from sql. 

This is a potential solution for https://github.com/kids-first/kf-lib-data-ingest/issues/627


fiendish:
clever

docs/source/tutorial/sourcing_data/query_sql.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Scenario
+========
+
+You would like to extract data directly from a PostgreSQL Database.
+
+.sql File Solution
+======================
+
+A convient way to load data from a postgres database is by saving your sql
```
gsantia, line 13:
```suggestion
A convenient way to load data from a postgres database is by saving your SQL
```

nicholasvk:
Very cool! Just want to chime in with some history/background toward developing standards on when it makes sense to use SQL directly. We thought about doing this for the CBTN refresh work which relies on data sources in the D3b warehouse. However, after discussions as a team with Allison/Bailey, we decided that we should write SQL output to Data Tracker via the study creator API for audit purposes and then use the files on data tracker for actual ingest. Database tables will change over time, rows will be added/deleted and/or updated, so having a record of exactly what is ingested for a given run seemed important at the time. A bonus is that Data Tracker makes this history easily accessible to non-ADAPT team members. You can review the code (authored by Meen and Avi) for that here:
https://github.com/d3b-center/d3b-warehouse-kids-first-refresh/blob/main/etl_from_eig_into_warehouse/dumping.py

Is this the best workflow? Are there other solutions to explore for this? Absolutely, just wanted to add to the conversation with history on the CBTN side.

chris-s-friedman:
@nicholasvk i think those are really good points about

1.  being able to audit what data is used for ingest
2. the volatility inherent to using sql as a data source - i.e. rows being added/ deleted/ updated
3. facilitate non-technical folks be able to have eyes on the data being used for ingest. 

my thinking proposing a tutorial on how to query sql in an ingest is for two specific circumstances:

1. ingesting information about genomic files from aws scrapes. instead of having ingesters generate aws manifests, we can pull directly from the file_metadata schema in postgres
2. provide a tutorial for external users that use the ingest library but don't have the data tracker but do use databases as source data. 

that said - I think I should prepend this tutorial with a note about your three points and how querying directly from sql impacts those points. Perhaps even say " instead of querying sql, you may want to create a static view of your database in a single file"

docs/source/tutorial/sourcing_data/query_sql.rst
```diff
@@ -0,0 +1,53 @@
+======================================================
+Source data stored in a PostgreSQL Database
+======================================================
+
+Scenario
+========
+
+You would like to extract data directly from a PostgreSQL Database.
+
```
chris-s-friedman, line 9: @nicholasvk how does this look?

```suggestion
.. note::
    Before deciding that you'd like to query a SQL database correctly, instead,
    you may want to consider separately querying your database, saving the
    result of that query to a file, and then using that file as source for an
    extract config. 

    By querying a database directly

        * It's no longer easy to audit the data used as the source for an
          ingest.
        * It's much harder, if not impossible, for non-technical users to know
          what data is used for an ingest
        * Using a database is inherently volatile e.g. rows may be added/
          deleted/ updated.

    Because of the above points, it's important to consider if your needs
    wouldn't be better served by using a static result of your query as
    source data for an ingest instead of directly querying the SQL database. 


```

nicholasvk: Looks great to me @chris-s-friedman . Thanks for considering my feedback and summarizing it nicely here!

gsantia (approved):

nicholasvk (approved):
Updated documentation for SQL considerations looks great!

PR #644, opened 2022-01-20
:sparkles: specify external IDs for clinical markers

chris-s-friedman:
add ability to specify external ids for diagnos, phenotype, and outcome. this is to enable the use of external ID when needed.


liberaliscomputing (approved):
Looks good!

PR #653, opened 2023-02-28
📝 add documentation on adding custom logs

chris-s-friedman:
adds documentation on how users can add user-defined logs to their extract configs and transform modules.


liberaliscomputing (approved):
You are good to go!

PR #654, opened 2023-03-06
👷‍♀️ Fix CI Errors

znatty22:
- **Fix lint error**: It's been a while since we've had a PR opened here. The lint check fails bc our `black` formatter finds linter errors. This must be because `black` got stricter with python style
-**Fix doc error**: todo
- **Fix tests error**: todo


chris-s-friedman (approved):

liberaliscomputing (approved):
✅ looks like fixing the downstream issues!

PR #663, opened 2023-06-14
♻️ Replace acl field with new authz

znatty22:
**DO NOT MERGE** until https://github.com/kids-first/kf-api-dataservice/pull/624 is in production

## Motivation

Authorization policy information for indexd files is currently stored in the `acl` list field. However, the `acl` field is being replaced with a new `authz` list field. The values in the `authz` list will correspond to policies created in [Arborist](https://github.com/uc-cdis/arborist). 

The following does not matter too much since the ingest library is not typically used to populate the `acl` field, but putting this here for some context:

- If the `authz` field is populated, any values in the `acl` field are ignored
- If the `authz` field is NOT populated, the values in the `acl` field are used as a fallback
- Because of the above two facts, we should manage the deprecated `acl` conservatively by setting it to an empty list (no access) by default

## Approach

- Add the new `authz` field to the KF dataservice target API plugin
- Set the value of `acl` field in KF dataservice target API plugin to be an empty list

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
 
   lint:
     docker:
-      - image: circleci/python:3.7
+      - image: circleci/python:3.9
     steps:
       - checkout
       - run: |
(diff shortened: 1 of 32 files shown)


chris-s-friedman (approved):
Looks good!

chris-s-friedman (approved):
Looks good....

For when we take this to a release, I'm wondering if we should consider that as a major release since this changes how previously written ingests will work if they ingest the ACL field...

PR #664, opened 2023-06-30
✨ create concepts for visibility reason and comment

chris-s-friedman:
Add visibility reason and comment, enumerated values for visibility reason, and adds both concepts to the kids first target api config.

See https://github.com/kids-first/kf-api-dataservice/pull/604 for the addition of these values into the kidsfirst dataservice


znatty22 (approved):
👍

PR #668, opened 2023-10-06
:heavy_plus_sign: add new constants for LRs

aw3334:
For long reads sequencing, the `se.library_selection` field should be set to either CCS or CLR.

- circular consensus sequencing (CCS)

- continuous long reads (CLR)

These values should be added as constants in the ingest library so that they will be accepted during long reads ingest. cc: @HigginsD3b

### DIFF
--- a/kf_lib_data_ingest/common/constants.py
+++ b/kf_lib_data_ingest/common/constants.py
@@ -301,6 +301,11 @@ class SEQUENCING:
         PANEL = "Panel"
         SCRNA = "scRNA-Seq"
         SNRNA = "snRNA-Seq"
+        CCS_WGS = "Circular Consensus Sequencing WGS"
+        CCS_RNA = "Circular Consensus Sequencing RNA-Seq"
+        CLR_WGS = "Continuous Long Reads WGS"
+        CLR_RNA = "Continuous Long Reads RNA-Seq"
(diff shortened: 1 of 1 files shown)


HigginsD3b (dismissed):
I'm not sure if the CLR and CCS long read sequencing methods are _exactly_ affecting library selection the way poly-T enrichment or rRNA depeletion do, but they do seem to be [different methods which affect the production of the sequencing library](https://ostr.ccr.cancer.gov/resources/sequencing-facility/faq-pacbio/#faq_9). I think the `se.library_selection` field may be the best for this information, as long as there aren't situations where any of the currently accepted short read values (eg: `PCR`, `Poly-T Enrichment`, `rRNA Depletion`) would also need to be applied to long read data labeled as `CLR` or `CLS`.

Hoping that @allisonheath or @yuankunzhu can weigh in here too.

znatty22 (dismissed):

HigginsD3b (approved):
[After discussion on Slack](https://d3b.slack.com/archives/CUXF6RA58/p1697122682825559), we agreed to move these values to `STRATEGY` as well as differentiate between WGS and RNA.

allisonheath (approved):
Looks good per discussion in slack to put LR types in experimental strategy.

PR #670, opened 2023-10-20
:heavy_plus_sign: adding ONT strategy

aw3334:
adding another value for long reads ingest that was previously missed

'ONT WGS'


Christina-J-Diaz (approved):
looks good!

PR #686, opened 2024-05-30
✨D3B-586 sample relationship

chris-s-friedman:


znatty22 (commented):
Looks good! But I think we need to modify the tests to include both sample and sample relationship. I forgot to make this comment on the sample PR

znatty22 (approved):

